PR #43, opened 2013-12-08
Refactored pam::service definition

tbartelmess:
The pam::service definition used to only have a free text 'content' parameter to define a pam service.
Now it takes an array of hashes for every type.

This breaks backwards compatibility, but I think this is a more sane way to define a pam service.


ghoneycutt:
@tbartelmess I love this though I'm a bit confused by it. Could you give some more examples on its usage, such as showing why you need an array of hashes?

ghoneycutt:
While I enjoy the array of hashes, I do not think that it actually simplifies things more than the array that is currently being used. It makes sense to me, but seems like more typing when you really just need an array of lines. Your validation function is particularly interesting to ensure that each line has the three necessary configuration elements. I cannot accept this PR, though would very much welcome the ability to validate an array of pam lines.

Issue #58, opened 2014-05-14
lsbdistmajrelease missing error on CentOS


knightsg:
Including the module gives an error about missing lsbdistmajrelease fact under CentOS (6.5 in this case). I included the redhat-lsb package in my manifest and added a "require" for this package to the pam class, however I still get the same fail message and the puppet run stops.

I installed redhat-lsb manually and re-ran puppet and it worked fine. For some reason, either puppet isn't installing redhat-lsb first, or puppet's logic hits the fail check in the pam module before it determines relationship ordering.

PR #67, opened 2014-08-11
Replace lsbmajdistrelease fact check with operatingsystemmajrelease for RedHat systems

treydock:
This relates to #58.  I do not like the dependencies pulled in by 'redhat-lsb' so using the `lsbmajdistrelease` fact is not an option in this case.  The `operatingsystemmajrelease` fact is simpler and requires no package dependencies.  AFAIK it's in all the stable versions of Facter (1.7 and 2.x).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -47,7 +47,7 @@ class pam (
 
   case $::osfamily {
     'RedHat': {
-      case $::lsbmajdistrelease {
+      case $::operatingsystemmajrelease {
         '5': {
           $default_pam_d_login_template = 'pam/login.el5.erb'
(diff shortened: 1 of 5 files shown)


splummer:
Any update on this pull request? Trying to use this on an OEL system and lsbmajdistrelease does not exist there.

ghoneycutt:
Could you please rebase this commit? It is no longer able to merge after we added EL7 support.

Once that's done, I'll be happy to merge.

ghoneycutt:
Rebased in PR #92

ghoneycutt:
Released in v2.12.0

PR #70, opened 2014-08-13
Add beaker acceptance tests

treydock:
Very basic beaker tests.  Adding other operating systems and operating system releases should only require adding new node definitions under `spec/acceptance/nodesets` and appropriate files under `spec/acceptance/support`.  May be useful to begin system based tested of future changes.

The file content checking isn't great, as I had issues making multiline strings work using:

``` ruby
its(:content) { should match expected_content }
```

Some of the file contents worked, some didn't.  Likely due to characters that break the matchers used in serverspec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ doc/
 coverage/
 spec/fixtures/modules/*
 Gemfile.lock
+
+# Beaker
+.vagrant/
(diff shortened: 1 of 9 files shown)


treydock:
This PR likely needs an update since beaker has progressed a lot since this was submitted and latest beaker does not support Ruby 1.8.7.

ghoneycutt:

PR #72, opened 2014-08-14
Add the 'purge_limits_d_dir' parameter to pam::limits

treydock:
Default is false for backwards compatibility.  Should probably be changed to true when a jump to version 3 is needed.


ghoneycutt:
Adding the new param to the README and making it work if it is stringified

ghoneycutt:
@treydock Thank you! Added a bit and tracking in PR #90

PR #73, opened 2014-09-05
Adding Ubuntu 14.04 Support

woodsbw:
Making changes needed to include support for Ubuntu 14.04 LTS


metadata.json
```diff
@@ -93,11 +94,11 @@
     "Gemfile.lock": "873e125e8afe971046172cf5e775871d",
     "LICENSE": "567e6cc3941d464d60c74d1f4abd5f46",
     "Modulefile": "8571add95d03e21ddcb6e3f90db325db",
-    "README.md": "b5281f1168d191d70f48b89f33c77629",
+    "README.md": "c82a61033650218f1d89b8fa55b236ee",
```
ghoneycutt: checksum junk is gone from metadata.json

manifests/init.pp
```diff
@@ -15,11 +15,13 @@
   $pam_d_login_group                   = 'root',
   $pam_d_login_mode                    = '0644',
   $pam_d_login_template                = undef,
+  $pam_d_login_template_content        = undef,
```
ghoneycutt: why the extra params for _content?

templates/login.ubuntu14.erb
```diff
@@ -0,0 +1,17 @@
+auth     optional   pam_faildelay.so  delay=3000000
```
ghoneycutt: This data should be in the manifest and the template should just iterate over the array from the manifest as the other templates do

ghoneycutt:
Hi @woodsbw

Made some suggestions to your PR. You are definitely on the right track, though tests are required for merging.

ghoneycutt:
@woodsbw Thanks for the PR, though going with #94 as it includes tests.

PR #74, opened 2014-09-18
EL 6 - Adding support for password-auth

nbentoumi:
Pam config for RHEL6 uses both system-auth and password-auth but the module only manages system-auth

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,41 @@ Content template of $system_auth_ac_file. If undef, parameter is set based on th
 
 - *Default*: undef, default is set based on OS version
 
+password_auth_file
+----------------
+Path to password-auth. Used on RedHat.
+
+- *Default*: '/etc/pam.d/password-auth'
+
+password_auth_ac_file
+-------------------
+Path to password-auth-ac. Used on RedHat.
+
+- *Default*: '/etc/pam.d/password-auth-ac'
+
+pam_password_auth_lines
+-------------------------
+Content template of $password_auth_ac_file. If undef, parameter is set based on the OS version.
+
+- *Default*: undef, default is set based on OS version
+
+pam_password_account_lines
+----------------------------
+Content template of $password_auth_ac_file. If undef, parameter is set based on the OS version.
+
+- *Default*: undef, default is set based on OS version
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -282,6 +282,41 @@ Content template of $system_auth_ac_file. If undef, parameter is set based on th
...
+----------------
+Path to password-auth. Used on RedHat.
+
+- *Default*: '/etc/pam.d/password-auth'
+
+password_auth_ac_file
+-------------------
+Path to password-auth-ac. Used on RedHat.
+
+- *Default*: '/etc/pam.d/password-auth-ac'
```
ghoneycutt: I don't see this file on EL6. Specifically CentOS 6.6

manifests/init.pp
```diff
@@ -36,6 +36,12 @@
   $common_session_noninteractive_file  = '/etc/pam.d/common-session-noninteractive',
   $system_auth_file                    = '/etc/pam.d/system-auth',
   $system_auth_ac_file                 = '/etc/pam.d/system-auth-ac',
+  $password_auth_file                  = '/etc/pam.d/password-auth',
```
ghoneycutt: all parameters need to be validated with corresponding spec tests

ghoneycutt:
Hi @nbentoumi 

Thanks for your contribution and apologies for such a late response. Would you be interested in working with me with to get this finished and merged? I need to understand what `/etc/pam.d/password-auth-ac` is, get the params validated and their corresponding spec tests added.

nbentoumi:
Hi @ghoneycutt 

Yes, we need to get this pull requested merged.
Pam config for RHEL6 uses both system-auth and password-auth but the module only manages system-auth.

"The problem with /etc/pam.d/system-auth is that it contains modules that are not usable in remote configurations so remote services such as sshd, vsftpd now use /etc/pam.d/password-auth." https://access.redhat.com/site/solutions/45492.

Thanks for help

ghoneycutt:
@nbentoumi will be happy to merge once it is ready with tests. Could you please answer the questions that are inline with the code?

nbentoumi:
Will do another pull request

PR #75, opened 2014-09-19
Add login_pam_access and sshd_pam_access parameters

treydock:
Add login_pam_access and sshd_pam_access parameters to override the control flag for pam_access.so when used in sshd.conf or login.conf templates (EL5, EL6 and SUSE 11)

This is particularly useful as I need to have pam_access set to "sufficient" in some cases when another PAM module is in charge of access control (pam_slurm.so).

I moved the tests for File[pam_d_login] and File[pam_d_sshd] resource to separate files for the sake of reducing the size of init_spec.rb.  Even with a good IDE it's sometimes hard to find things in a file over 1500 lines long.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,6 +4,8 @@
 #
 class pam (
   $allowed_users                       = 'root',
+  $login_pam_access                    = 'required',
+  $sshd_pam_access                     = 'required',
   $ensure_vas                          = 'absent',
   $package_name                        = undef,
   $pam_conf_file                       = '/etc/pam.conf',
@@ -754,6 +756,11 @@ class pam (
       fail("Pam is only supported on RedHat, Suse, Debian and Solaris osfamilies. Your osfamily is identified as <${::osfamily}>.")
     }
   }
+
+  $valid_pam_access_values = ['^required$', '^requisite$', '^sufficient$', '^optional$', '^absent$']
+
+  validate_re($login_pam_access, $valid_pam_access_values)
+  validate_re($sshd_pam_access, $valid_pam_access_values)
 
   if $package_name == undef {
     $my_package_name = $default_package_name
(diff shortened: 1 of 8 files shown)


ghoneycutt:
@treydock The cleanup should be in a different PR than adding this functionality. I want the functionality, but do not understand all of the changes you made, also since they dramatically change the spec file, it makes rebasing and merging very difficult.

treydock:
@ghoneycutt The cleanup was really unnecessary but something I did so I could better follow the tests since the one spec file was very long.  I've started using rspec-puppet-facts, https://github.com/mcanevet/rspec-puppet-facts, which drastically simplifies module testing against numerous Operating Systems and versions.  Not sure if it's applicable here but worth a look.

I'll update PR with spec refactor removed.

treydock:
Rebased against latest changes and maintained same spec layout.  I'll have to dig in my fork to find where I added the pam_access.so line to EL7 systems as I noticed my few EL7 servers are using pam_access.so which isn't part of this PR.

treydock:
@esalberg That's a good point.  If @ghoneycutt wants such a sanity check performed I'll see what it'll take to make that happen.

ghoneycutt:
Thanks for coming back and taking a look at this, I really appreciate it!


rspec-puppet-facts looks pretty cool. There is a similar approach here[1] that is used in some of the modules. As a bit of history, this module's rspec tests were solely written by people who were learning rspec-puppet at the time and the module just kept growing and growing to support more OS's. Definitely in agreement that the spec file could be condensed quite a bit.

I think I prefer the platforms hash method from the vim module because there is less magic on what is going on, you can supply other data such as package names, and because rspec-puppet-facts does not support the Solaris and Suse versions that this module supports.

[1] - https://github.com/ghoneycutt/puppet-module-vim/blob/master/spec/classes/init_spec.rb#L4-148

ghoneycutt:
This commit rocks! I'll go ahead and document the params in the README. While not Puppet Labs styling, I document every input in the README along with default values so people know what they can tweak in Hiera without having to go through the code.

ghoneycutt:
Added bits to the README, error messages for the validation, and some tests in PR #93

ghoneycutt:
Released in v2.13.0

PR #76, opened 2014-09-19
treydock's branch

ghoneycutt:


spec/classes/accesslogin_spec.rb
```diff
@@ -22,26 +22,19 @@
...
-
-# allow only the groups listed
-+ : root : ALL
-
-# default deny
-- : ALL : ALL
-})
-      }
+      it "should contain default allowed_users" do
+        verify_contents(subject, 'access_conf', [
```
ghoneycutt: ?

manifests/init.pp
```diff
@@ -47,7 +51,7 @@
 
   case $::osfamily {
     'RedHat': {
-      case $::lsbmajdistrelease {
+      case $::operatingsystemmajrelease {
```
ghoneycutt: This can work in this use case since it is under RedHat. If used elsewhere it will break as Suse does not support the new fact.

ghoneycutt:
@treydock your work is phenomenal! It is just too much all in one commit to just merge in. If you could split new functionality or bug fixes into their own pull requests, it is much easier to work together to get things merged. As it is now, I would need to cherry pick all the awesome around the few things I cannot merge or just do not understand enough to merge.

treydock:
@ghoneycutt No problem, I'll try and make individual branches for each change and open up a PR per branch.

treydock:
@ghoneycutt Some of my changes have been rushed and not very good for a module used by the community.  Do you see any changes I've made that would benefit this module that haven't already been merged?  I know my method of unit testing is different, but for actual functionality of the module I'm having trouble identifying what else could be of use.

ghoneycutt:

PR #82, opened 2014-09-30
Add ensure_vas for Solaris 11

juliengk:
Last one for Solaris for now :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -9,12 +9,22 @@ class pam (
   $pam_conf_file                       = '/etc/pam.conf',
   $services                            = undef,
   $limits_fragments                    = undef,
+  $pam_d_cron_path                     = '/etc/pam.d/cron',
+  $pam_d_cron_owner                    = 'root',
+  $pam_d_cron_group                    = 'root',
+  $pam_d_cron_mode                     = '0644',
+  $pam_d_cron_template                 = undef,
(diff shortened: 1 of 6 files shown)


ghoneycutt:
@juliengk looks like comments were lost during rebasing. Could you have someone verify that this works for them as I'm unable to test Solaris 11.

manifests/init.pp
```diff
@@ -9,12 +9,22 @@
   $pam_conf_file                       = '/etc/pam.conf',
   $services                            = undef,
   $limits_fragments                    = undef,
+  $pam_d_cron_path                     = '/etc/pam.d/cron',
```
ghoneycutt: All new parameters need to be validated.

templates/cron.sol11.erb
```diff
@@ -0,0 +1,8 @@
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+account definitive        pam_user_policy.so.1
+<% if @ensure_vas == 'present' -%>
```
ghoneycutt: why are there variable here in the template when there is already conditional logic in the manifest to set the contents of this file?

These should look like the other templates.

ghoneycutt:
@Phil-Friderici and @emahags contributed to a huge change to simplify spec tests and now this PR is out of date. If you are still interested, you will need to rebase against master.

PR #84, opened 2014-11-28
Update init.pp

pmiles:
In our organisation, we don't need all of the functionality of the puppet nsswitch module.

This change wraps a condition around the nsswitch include, and if nsswitch module doesn't exist, silently ignore the include instead.

I believe this change is backward compatible.

ptionally include nsswitch module if its installed.


ghoneycutt:
Closing this due to inactivity. Opened issue #89 with an example, should you want to pursue this.

PR #88, opened 2015-01-28
Epic change in array layout reduces excessive line length that hurts my editor

Aethylred:
More compact, easier to read and overall lower sanity check threshold.

No code changed, just whitespace.


ghoneycutt:
@Aethylred I really appreciate this, though I'm reticent to merge it because it will make it impossible for anyone else to cleanly merge their code. I need to check the open PR's to see what is viable.

ghoneycutt:
Released in v2.11.0

PR #91, opened 2015-01-28
Order hash for access.conf, useful for Ruby 1.8.7

ghoneycutt:
Thanks to Treydock for pointing this out! Prior to this commit, if you
specify allowed_users as a hash with Ruby 1.8.7 it would change the
order and show your puppet runs as having changes.


ghoneycutt:
Released in v2.11.0

PR #92, opened 2015-01-28
Replace lsbmajdistrelease fact check with operatingsystemmajrelease for ...

ghoneycutt:
...RedHat based systems.

The use of lsbmajdistrelease must remain for Suse until the operatingsystemmajrelease fact supports Suse.


ghoneycutt:
Released in v2.12.0

PR #95, opened 2015-01-30
Add SLES10.x support

propyless:
SLES10.x systems dont look in the limits.d catalog and therefore fragments don't work.

I've modified the module so that for 10.x systems the limits are inserted into the limits.conf file instead.


propyless:
will send a pr later without conflicts......

ghoneycutt:
@propyless You will want to add me as a remote 'upstream' and do `git pull upstream master` then create your branch to work. This will make merging easier.

PR #98, opened 2015-02-04
Send notice instead of hard fail

propyless:
thought it might be better to send a notice rather than fail the whole puppet run.
this is just to see if you think this might be better.

//propy

### DIFF
--- a/manifests/limits/fragment.pp
+++ b/manifests/limits/fragment.pp
@@ -10,36 +10,36 @@ define pam::limits::fragment (
   include pam
   include pam::limits
   if $::osfamily == 'Suse' and $::lsbmajdistrelease == '10' {
-    fail('You can not use pam::limits::fragment together with Suse 10.x releases')
-  }
-
-  # must specify source or list
-  if $source == 'UNSET' and $list == undef {
-    fail('pam::limits::fragment must specify source or list.')
-  }
-
-  # list takes priority if you specify both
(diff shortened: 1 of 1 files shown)


manifests/limits/fragment.pp
```diff
@@ -10,36 +10,36 @@
   include pam
   include pam::limits
   if $::osfamily == 'Suse' and $::lsbmajdistrelease == '10' {
-    fail('You can not use pam::limits::fragment together with Suse 10.x releases')
```
ghoneycutt: defines/classes should be explicit about what they support and if you use an unsupported system, they should fail.

propyless:
Alright yeah thats what I thought as well. If it the OS does not support pam fragments, the puppet run should fail imo. We had some mixed opinions about it failing here at work and I thought i'd provide an alternate solution.

ghoneycutt:
Thanks for asking. Always better to fail the catalog when unsure of compatibility as opposed to doing something that may not work as intended. If you checkout out the pam class you'll see the osfamily case statement where the defaults are always to fail.

propyless:
There opinion derives from the fact that people (including us) have been using fragments and sle 10.x and now we would receive fails.

Phil-Friderici:
thanks for your thoughts :)

propyless:
closed because it should fail.

PR #100, opened 2015-02-09
Make pam::service resources reversible

Aethylred:
What can be locked with puppet, can now be unlocked with puppet.

Previously if you defined a pam::service resource and then removed it, the content would remain on the node. This could leave vital services (e.g. `sudo`) locked and unusable. This PR does the following:

* adds an `ensure` parameter to `pam::service` to actively remove resources created on the node when set to `absent`
* adds a `lines` parameter to `pam::service` so it can be handed an array of strings like the other resources in the module.
* Includes spec tests for `pam::service` which were not done at all previously.


README.md
```diff
@@ -419,7 +419,12 @@ pam::services:
     content : "auth     required       pam_unix2.so"
 </pre>
 
-## Paramteters for `pam::service`
+## Parameters for `pam::service`
+
+ensure
+------
+
+Specifies if a PAM service file should (`present`) or should not (`absent`) exist.
```
ghoneycutt: This should document the default setting of present like the other params.

README.md
```diff
@@ -431,6 +436,10 @@ content
 -------
 Content of the PAM file for the service
 
+lines
```
ghoneycutt: Same here. This should also mention that it is mutually exclusive with content and that content or lines is required to be set if ensure is present.

manifests/service.pp
```diff
@@ -3,16 +3,39 @@
...
 define pam::service (
+  $ensure         = 'present',
   $pam_config_dir = '/etc/pam.d',
   $content        = undef,
+  $lines          = undef
 ) {
 
   include pam
 
+  validate_re($ensure, ['^present$', '^absent$'] )
```
ghoneycutt: Does this work if it is set to `absent` instead of `'absent'` ?

spec/defines/service_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+          {
+            :lines  => [
+              '@include common-auth',
+              '@include common-account',
+              'session required pam_permit.so',
+              'session required pam_limits.so'
+            ]
+          }
+        end
+      it { should contain_file('pam.d-service-test').with_content(
```
ghoneycutt: Since the ordering of the lines is important to PAM we should check that the entire file's contents match instead of per line checks. Here's an example https://github.com/ghoneycutt/puppet-module-dnsclient/blob/master/spec/classes/init_spec.rb#L18-26

templates/service.erb
```diff
@@ -0,0 +1,5 @@
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+<% @lines.each do |line| -%>
+<%= line %>
+<% end -%>
```
ghoneycutt: There should be a newline at the end of the file.

ghoneycutt:
Awesome work @Aethylred !

Aethylred:
Will get onto this tomorrow, because Kidney stone.

I've got another PR incoming that handles some `libpam` modules.

README.md
Aethylred: Neither the lines or content parameters are required, not setting either will result in an empty service file... which is replicating the behavior of not setting the content parameter before this patch.

manifests/service.pp
Aethylred: yes, though bare strings are being depreciated with the Future Parser and sooner or later will start throwing syntax validation warnings.

spec/defines/service_spec.rb
Aethylred: I had been wondering how to do that (even if it is ugly).

Aethylred:
Ready to go!

spec/defines/service_spec.rb
ghoneycutt: And it certainly is ugly :>

ghoneycutt:
Thanks @Aethylred 

Squashed your commits into one and did a minor modification to the spec tests in PR #103

PR #101, opened 2015-02-12
`pam::lib` classes to manage PAM libraries

Aethylred:
A number of `pam::lib`  classes to install some of the PAM libraries that allow PAM to use different authentication protocols.

They're basically wrappers for a package declaration that would tie into an appropriately configured service or client (which puts those outside the scope of this module).


Aethylred:
This is good to go too.

manifests/lib/ccreds.pp
```diff
@@ -0,0 +1,34 @@
...
+        fail("a custom PAM ccreds module package is required for ${::osfamily}")
+      }
+    }
+  }
+
+  package{$my_package:
+    ensure => $ensure,
+  }
+
+}
```
ghoneycutt: missing new line. perhaps this is configurable in your editor?

ghoneycutt:
The OS's should have subcases for the versions that we support, that we are explicit about the package name to platform mapping. @Phil-Friderici might be able to help with OS's to which you might not have access.

Tests look good, though for the myriad of OS/version choices, suggest using this approach that defines a hash of platforms and their associated packages

https://github.com/ghoneycutt/puppet-module-vim/blob/master/spec/classes/init_spec.rb#L4-122

Phil-Friderici:
Please ignore my ignorance: does it make sense to put these one-resourcers into own subclasese ?
If this would be in the main class, we could also auto-add the needed lines (pam::service::lines).

ghoneycutt:
I think @Aethylred wants to include these in other bits of code. A kerberos module might include pam::lib::krb5 for instance in addition to the rest of the kerberos configs.

Phil-Friderici:
hmmm let the pam module install kerberos packages for usage in other modules ?
I would use [types::package] (https://github.com/ghoneycutt/puppet-module-types) instead.

Anyway, if you support this I will do my best to help :)

Aethylred:
I intend these just to figure out which is the appropriate package to install on each OS to enable that kind of authentication, and to have them installed in the right sequence to make it work. Having these auto add lines into services or login could get messy, as it needs to be tunable to work with whatever policy-of-the-day the infosec team convinces management to implement. E.g kerberos to login, and RADIUS for sudo.

ghoneycutt:
This will be good to go once the osfamily case statements have sub conditions for the versions. Not having those makes a lot of assumptions and even if they work for all versions that we currently support, newer OS versions can inadvertently break this. Just saw this happen with puppetlabs/ruby and EL7.

@Aethylred If you could please add those and for any OS to which you do not have access, just leave it out of the PR. Other folks using those OS's can add onto it to fill those in.

Aethylred:
@ghoneycutt packages are correct for Debian and Ubuntu (they're the same between distros and across versions.

They also seem to be consistent in RHEL based distros since Fedora 7, which includes OpenSuSE: http://rpm.pbone.net/index.php3?stat=3&limit=1&srodzaj=1&dl=40&search=pam_ccreds&field[]=1&field[]=2

ghoneycutt:
That link is an awesome resource!

ghoneycutt:
While they are the same in the past versions, I still think we need to be explicit in the case statement so that we are not burned on future versions. Something like the snippet below. Would you like me to build on top of your PR or do you want to handle it?

``` puppet
case $::osfamily {
  'RedHat': {
    case $::operatingsystemmajrelease {
      '5','6','7': {
        $default_package = 'pam_ccreds'
      }
      default: {
        fail()
      }
    }
  default: {
    fail()
  }
}
```

Aethylred:
I should be able to put that together tomorrow.

Aethylred:
Travis in maintenance. may need to restart tests when It's back.

ghoneycutt:
Going through the spec tests.

ghoneycutt:
@Aethylred Sent a PR your way on this branch.

https://github.com/Aethylred/puppet-module-pam/pull/2

Reworked the manifest a bit to be in line with the rest of the code and redid the spec tests entirely. Could you use this as a pattern for the other three classes?

Aethylred:
Will check them out.

Aethylred:
Finally! :weary: 

I've checked these out and it looks to be good. I'm trialing them in a new deployment and seems to be functionally identical. Tests look good too, but waiting for Travis to queue them up...

Aethylred:
Now I have to figure out who I am and what I was doing here... and what needs to be done :P

Aethylred:
rebased and cutpasted.

Aethylred:
...and still works :)

ghoneycutt:
Wow, 8 months behind :( Apologies for the lag.

@Phil-Friderici what do you think about this?

Phil-Friderici:
Code wise it looks good and complete to me.

*pam::lib::ccreds does not have a default for your system. Please specify a package at pam::lib::ccreds::package.*
should be changed as it will also fail when you specify the package.

@Aethylred Interesting approach for variable validations. Maybe you like to have a look on my quite flexible variant [1] Comments are always welcome.

[1] https://github.com/Phil-Friderici/puppet-rngd/blob/master/spec/classes/init_spec.rb#L130-L215

ghoneycutt:

PR #102, opened 2015-02-17
Don't configure pam if vas installation failed

emahags:
Configuring pam to use vas when the vas installation failed will cause
the machine to enter a broken state where logins are not possible.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1263,4 +1263,18 @@ class pam (
       fail("Pam is only supported on RedHat, SuSE, Debian and Solaris osfamilies. Your osfamily is identified as <${::osfamily}>.")
     }
   }
+
+  if $ensure_vas == 'present' and ($::osfamily == 'RedHat' or $::osfamily == 'Debian' or ($::osfamily == 'Suse' and $::lsbmajdistrelease != '9')) {
+    include vas
+    File <| title == 'pam_common_auth'                   |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_auth_pc'                |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_account'                |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_account_pc'             |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_password'               |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_password_pc'            |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_noninteractive_session' |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_session'                |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_common_session_pc'             |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+    File <| title == 'pam_system_auth_ac'                |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
+  }
 }
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -9,5 +9,14 @@ fixtures:
     common:
       repo: 'git://github.com/ghoneycutt/puppet-module-common.git'
       ref: 'v1.0.2'
+    vas:
+      repo: 'git://github.com/Ericsson/puppet-module-vas.git'
+      ref: 'v0.10.0'
+    nisclient:
+      repo: 'git://github.com/Ericsson/puppet-module-nisclient.git'
+      ref: 'v1.0.0'
+    rpcbind:
+      repo: 'git://github.com/ghoneycutt/puppet-module-rpcbind.git'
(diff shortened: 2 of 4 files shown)


manifests/init.pp
```diff
@@ -938,14 +938,26 @@
...
 
-          file { 'pam_system_auth_ac':
-            ensure  => file,
-            path    => $system_auth_ac_file,
-            content => template('pam/system-auth-ac.erb'),
-            owner   => 'root',
-            group   => 'root',
-            mode    => '0644',
-            require => Package[$my_package_name],
+          if $ensure_vas == 'present' {
```
ghoneycutt: inside the if statement use the reference syntax so the code is not duplicated such as the following.

``` puppet
File['pam_system_auth_ac'] {
  require => [ Package[$my_package_name], Exec[vasinst], ],
}
```

ghoneycutt:
Will need to update spec tests to have a context for with vas and without vas to ensure that the require attribute is set appropriately on those files.

Ugh. Looking at the spec tests and it is way past due to move to a hashed system like vim

https://github.com/ghoneycutt/puppet-module-vim/blob/master/spec/classes/init_spec.rb#L4-60

emahags:
@ghoneycutt Didn't know you could do it like that, it makes it so much simpler. Thanks! I had to use this kind of syntax though as I got redefine errors otherwise:
```
File <| title == 'pam_common_auth' |> { require => [ Package[$my_package_name], Exec[vasinst], ], }
```
Should I wait for a rewrite of the spec tests or try getting the hashed system in for this?

ghoneycutt:
If you want to take on refactoring the spec tests that would be awesome!! That should be in a different PR that doesn't add any functionality though.

Up to you if you want to do this one first and add the spec tests or do the refactor and then add this functionality after that.

ghoneycutt:
Hmm. You shouldn't have to use the spaceship collector for this as the reference syntax should work.

emahags:
@ghoneycutt I got this error when I did that:
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Parameter 'require' is already set on File[pam_system_auth_ac] by #<Puppet::Resource::Type:0x5dfb7e3f> at /etc/puppet/environments/vaspam/
modules/pam/manifests/init.pp:948; cannot redefine at 1302:/etc/puppet/environments/vaspam/modules/pam/manifests/init.pp
```
Maybe I did it in the wrong way. Should I try getting it to work with references or are spaceships fine?

I'll give it a go with the refactoring in another PR.

ghoneycutt:
Here's an example

https://github.com/puppetlabs/puppetlabs-concat/blob/eda6f2cc6e76f7e9f65c0704cca43df82498bfe2/manifests/init.pp#L192-197

manifests/init.pp
```diff
@@ -1263,4 +1263,19 @@
       fail("Pam is only supported on RedHat, SuSE, Debian and Solaris osfamilies. Your osfamily is identified as <${::osfamily}>.")
     }
   }
+
+  if $ensure_vas == 'present' {
```
ghoneycutt: Exec['vasinst'] is somewhere in the vas module. You need to include that class here to ensure that the dependency is found. This would also mean updating the README. The metadata does not need to be updated, since the vas module is not a hard requirement, though the README should not that if you want that functionality, you need Ericsson/vas >= x.y.z

ghoneycutt:
Looks like this just needs the README and spec tests modified to be able to be merged.

emahags:
@ghoneycutt The readme didn't contain ensure_vas to begin with, I've added it now.

manifests/init.pp
```diff
@@ -1263,4 +1263,18 @@
       fail("Pam is only supported on RedHat, SuSE, Debian and Solaris osfamilies. Your osfamily is identified as <${::osfamily}>.")
     }
   }
+
+  if $ensure_vas == 'present' and ($::osfamily == 'RedHat' or $::osfamily == 'Debian' or ($::osfamily == 'Suse' and $::lsbmajdistrelease != '9')) {
+    include vas
```
ghoneycutt: vas already includes pam, though it does not appear to reference any of the resources within pam, so this should be safe from preventing circular dependencies. Have you confirmed in your environment by running this code on a new system?

manifests/init.pp
```diff
@@ -1263,4 +1263,18 @@
       fail("Pam is only supported on RedHat, SuSE, Debian and Solaris osfamilies. Your osfamily is identified as <${::osfamily}>.")
     }
   }
+
+  if $ensure_vas == 'present' and ($::osfamily == 'RedHat' or $::osfamily == 'Debian' or ($::osfamily == 'Suse' and $::lsbmajdistrelease != '9')) {
```
ghoneycutt: if ensure_vas is present on an unsupported platform, shouldnt it fail so that you know your configuration is not valid. This would seem to silently skip it.

emahags: @ghoneycutt That would lock out people configuring pam to use vas on those platforms until support is added to the vas module though. I'm not sure which is better.

manifests/init.pp
emahags: @ghoneycutt I have run this on a few systems without any problems.

ghoneycutt:
@emahags @Phil-Friderici Not sure if this is still needed. What's the status here?

emahags:
@ghoneycutt  This functionality was recently merged in the vas module. Thanks for reminding me.

PR #109, opened 2015-03-31
Add ensure attribute for limits fragment

jwennerberg:


ghoneycutt:
Released in v2.16.0

Thanks @jwennerberg and @Phil-Friderici

PR #110, opened 2015-04-13
Add support for OpenSuSE 13.1.

buzzdeee:
Updated README.md, metadata.json, and added spec tests.

The default values are taken from a fresh installed box, with local authentication.

I don't really know what this 'vas' is good for, and I could not test it, so the module
fails when trying to enable vas on OpenSuse 13. Getting the specs right for that
was a bit tricky, but I check if it's suse 13, and then jump to the next iteration of the
specs loop.

let me know if there is something that needs changes/enhancements....

cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ This module has been tested to work on the following systems using Puppet v3 wit
  * Suse 10
  * Suse 11
  * Suse 12
+ * OpenSuSE 13.1
  * Ubuntu 12.04 LTS
  * Ubuntu 14.04 LTS
 
--- a/metadata.json
+++ b/metadata.json
@@ -83,6 +83,12 @@
       ]
     },
     {
+      "operatingsystem": "OpenSuSE",
+      "operatingsystemrelease": [
+        "13.1"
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -22,6 +22,7 @@ This module has been tested to work on the following systems using Puppet v3 wit
  * Suse 10
  * Suse 11
  * Suse 12
+ * OpenSuSE 13.1
```
ghoneycutt: Is `Suse 13` accurate?

metadata.json
```diff
@@ -83,6 +83,12 @@
       ]
     },
     {
+      "operatingsystem": "OpenSuSE",
```
ghoneycutt: As per your other PR, I'm not sure this metadata is correct. I'll investigate some more.

ghoneycutt:
Awesome work!

VAS aka QAS is Dell's Quest Authentication Services.

ghoneycutt:
Thanks @buzzdeee 

Released in v2.19.0

PR #113, opened 2015-06-08
Solaris11 support

juliengk:


manifests/init.pp
```diff
@@ -1292,6 +1339,33 @@
           }
         }
         '5.11': {
+          file { 'pam_d_cron':
```
ghoneycutt: Curious why these files were not already in the module since the module already supposedly supports Solaris 11

juliengk: good question!

ghoneycutt:
Looks like @kentjohansson originally added support for Solaris 11. Since I don't run that platform, we'll just need confirmation from another source/hub that these changes are OK. We want the default values for Solaris 11, not the Ericsson defaults.

juliengk:
I will ask another hub to test it once I will be ready. Thanks

Phil-Friderici:
This version was successfully tested by a co-worker, he wrote:
> I have tested the pam module and now works ok on Solaris 11.1 and fails on Solaris 11.0 – as expected.

Guess it's time to fix the spec tests now...

PR #114, opened 2015-06-09
Add the ability to use System Security Services Daemon (SSSD) for authentication

herlo:
When using FreeIPA, (and LDAP, too), SSS provides the functionality for sudo via pam. This patch enables a default for a flag 'enable_sss'. Additiionally, I removed 'auth sufficient pam_fprintd.so' as this is not standard on server systems. Creating a config set of lines would be more appropriate.

access.conf should use the defaults if nothing is set. However, because of the way it is coded, a hiera value "pam::allowed_users: 'ALL'" is needed to adjust this to ensure default file would be in place. If no hiera exists, the original behavior of root only being allowed will occur.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -7,6 +7,7 @@ class pam (
   $login_pam_access                    = 'required',
   $sshd_pam_access                     = 'required',
   $ensure_vas                          = 'absent',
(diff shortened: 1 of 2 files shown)


herlo:
adding additional fixes.

PR #115, opened 2015-06-09
Add the ability to use System Security Services Daemon (SSSD) for authentication

herlo:
When using FreeIPA, (and LDAP, too), SSS provides the functionality for sudo via pam. This patch enables a default for a flag 'enable_sss'. Additiionally, I removed 'auth sufficient pam_fprintd.so' as this is not standard on server systems. Creating a config set of lines would be more appropriate.

access.conf should use the defaults if nothing is set. However, because of the way it is coded, a hiera value "pam::allowed_users: 'ALL'" is needed to adjust this to ensure default file would be in place. If no hiera exists, the original behavior of root only being allowed will occur.

### DIFF
--- a/templates/access.conf.erb
+++ b/templates/access.conf.erb
@@ -16,10 +16,14 @@ entries = scope.lookupvar('pam::allowed_users')
 + : <%= key %> : ALL
 <% end -%>
 <% elsif entries.is_a? String -%>
+<% if entries != 'ALL' -%>
 + : <%= entries %> : ALL
-<% else -%>
-+ : root : ALL
-<% end -%>
-
 # default deny
 - : ALL : ALL
+<% else -%>
++ : root : ALL
+# default deny
+- : ALL : ALL
+<% end -%>
+<% end -%>
+
--- a/templates/access.conf.erb
+++ b/templates/access.conf.erb
@@ -20,10 +20,10 @@ entries = scope.lookupvar('pam::allowed_users')
 + : <%= entries %> : ALL
 # default deny
 - : ALL : ALL
+<% end -%>
 <% else -%>
 + : root : ALL
 # default deny
 - : ALL : ALL
-<% end -%>
 <% end -%>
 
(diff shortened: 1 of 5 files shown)


templates/access.conf.erb
```diff
@@ -16,10 +16,14 @@ entries = scope.lookupvar('pam::allowed_users')
...
+- : ALL : ALL
+<% end -%>
 <% else -%>
 + : root : ALL
-<% end -%>
-
 # default deny
 - : ALL : ALL
+<% end -%>
+
```
ghoneycutt: no newline here

ghoneycutt:
Hi @herlo 

Appreciate the work!!

The addition of `ensure_vas` caused the need for a ton more code and complexity in providing defaults and the surrounding spec tests. I think this is going to do the same and it only supports one OS, EL7. I think it would be more useful to simply document the settings needed to get SSS working with the module on EL7. What do you think?

Could you elaborate on why you do not think that `auth sufficient pam_fprintd.so` is appropriate? I would love to learn more. Not sure what you mean by "server" systems either. This module is used on all types of systems and some you might not think of as servers, such as my laptop :)

Tip: When submitting PR's to projects, group them by specific bug fixes or features. In this PR you tried both and on three different topics, which makes it harder to be merged. Smaller PR's will get you faster merges.

herlo:
Hi @ghoneycutt,

I realize I had combined the patches together. I couldn't figure out a way to do a PR for just one or two at a time. I'll look into fixing that in the future.

As for the fprintd.so component. The reason I think it's not appropriate is that the default varies per system. Most server systems, meaning ones that likely run in a rack in a datacenter would never use fprintd.so. I don't have a problem with fprintd.so for systems that have a fingerprint reader, but the value is lost as a default since many (most?) systems that would use this code would not. The thought process was that a user could always add it back with the 'lines' option.

SSS actually works on el5, el6, el7, ubuntu, debian and much, much more. I only was able to test it for el7 at this time, however, so I didn't want to assume responsibility for making it work on all of these other systems.

Hope this helps explain why I did what I did. 

Cheers,
herlo

ghoneycutt:
Appreciate the work, though I cannot accept this for the reasons above.

I would greatly appreciate a pull request to the README that shows what needs to be set in Hiera to enable this to work on EL7.

PR #116, opened 2015-06-24
Set pam_access on EL7 for sshd and login - treat the same as EL5 and EL6

treydock:
This is a change I've made in my fork (ref #76) for a while and have branched it off for acceptance.  This basically treats EL7 just like EL5 and EL6 in terms of the inclusion of pam_access.

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -435,7 +435,7 @@ describe 'pam' do
         end
         let(:params) {{ :login_pam_access => 'sufficient' }}
 
-        if (v[:osfamily] == 'RedHat' and (v[:release] == '5' or v[:release] == '6')) or (v[:osfamily] == 'Suse' and v[:release] == '11')
+        if (v[:osfamily] == 'RedHat') or (v[:osfamily] == 'Suse' and v[:release] == '11')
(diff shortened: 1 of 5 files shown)


ghoneycutt:
Is this the default for EL7 or would this change a stock configuration?

treydock:
@ghoneycutt The default in all EL systems (EL5-7) is not to use pam_access.  At least this is true of a minimal install.  I haven't done a non-Puppet install in a long time, but am pretty sure the default value for `login_pam_access` and `sshd_pam_access` do not conform to a stock configuration.

This pull request is more just to make the module's capabilities for EL7 match that of EL5-6.  Making things truly match a stock config (which is a good idea) would require backwards incompatible changes to default parameter values.

ghoneycutt:

Phil-Friderici:
Sorry, I don't have enough knowledge about pam details to have an opinion on the functionality.
Looking at the code, I would like to ask for spec testing the added functionality.

treydock:
I've updated the PR to test the changes.  The tests already existed, I just extended them to also include EL7.

ghoneycutt:
Thank you @treydock !

Released in v2.29.1

Without this patch, pam::allowed_users which manages access.conf would not work.

PR #118, opened 2015-07-10
Ability to turn off management of accesslogin and limits files

edestecd:
I want to manage these with Augeas

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,11 +44,14 @@ class pam (
   $system_auth_ac_password_lines       = undef,
   $system_auth_ac_session_lines        = undef,
   $vas_major_version                   = '4',
+  $manage_nsswitch                     = true,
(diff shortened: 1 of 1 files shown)


ghoneycutt:
@edestecd These changes look good. In order to be merged all attributes need to be documented in the README and there must be corresponding spec tests.

The spec tests should show that the default behavior works as expected, that when they accept values  the expected thing happens and that if you pass an incorrect value that an error occurs.

In concrete terms the defaults should be added to have

``` ruby
  it { should contain_class('foo') }
```

and if you set them to false

``` ruby
  it { should_not contain_class('foo') }
```

and if they are incorrect it should Error

``` ruby
it 'should fail' do
  expect {
    should contain_class('foo')
  }.to raise_error(Puppet::Error,/put_error_message_here/)
end
```

Let me know if you need any other help for the spec tests.

PR #121, opened 2015-07-29
Removed unnecessary inclusion of class 'pam' from 'pam::service'

nomadcz:


ghoneycutt:
That is necessary as the pam class installs the pam packages and ensures the directories needed in the service define exist.

Please re-open if you feel I'm missing something or if the current behavior is causing a problem for you as opposed to an optimization.

PR #122, opened 2015-10-26
Solaris 11.1 support

Phil-Friderici:
based on @juliengk https://github.com/ghoneycutt/puppet-module-pam/pull/113

Fixed the spec tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This module has been tested to work on the following systems using Puppet v3 wit
  * EL 7
  * Solaris 9
  * Solaris 10
- * Solaris 11
+ * Solaris 11.1
  * Suse 9
  * Suse 10
  * Suse 11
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ This module has been tested to work on the following systems using Puppet v3 wit
  * EL 7
  * Solaris 9
  * Solaris 10
- * Solaris 11
+ * Solaris 11.1
```
ghoneycutt: Is 11.1 different than 11 ?

ghoneycutt:
It appears we are dropping support for 11 while adding support for 11.1. Could you explain?

README.md
Phil-Friderici: got this information from @nalyanyam while he tested this code
> using the Solaris 11.1 configuration on 11.0 had real bad consequences – (such as login
> without password, passwd command fail etc).

Phil-Friderici:
got this information from @nalyanyam while he tested this code
> However, there is a difference in pam configuration files between earlier version of Solaris - 11.0
> and Solaris 11.1 Solaris 11.0 has same configuration as earlier OS kernelrealses such as Solaris 10
> – meaning it requires /etc/pam.conf. While beginning form Solaris 11.1, there is an alternative > configuration that uses /etc/pam.d/<service> design

@juliengk could you please explain why Solaris 11.0 now fails when it could be treated like Solaris 10 ?

nalyanyam:
@juliengk, Probably you could update the module such that when the Kernelversion is 11.0, it uses same pam configuration as Solaris 10. Should be similar but i will need to test it again after you have done the changes. 

PAM configuration via the per-service PAM policy files in the /etc/pam.d directory in addition to the pam.conf file, is supported starting from Solaris 11.1

juliengk:
@Phil-Friderici i don't have time that's why!

ghoneycutt:
Closing due to lack of response. Please feel free to re-open.

PR #125, opened 2016-01-12
Added support for Debian Jessie (8)

jplana:
This PR adds basic support for Debian current stable (Jessie at the moment), which is quite similar to Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ only), 1.9.3, 2.0.0 and 2.1.0.
  * OpenSuSE 13.1
  * Ubuntu 12.04 LTS
  * Ubuntu 14.04 LTS
+ * Debian 8.2
 
 EL no longer requires the `redhat-lsb` package.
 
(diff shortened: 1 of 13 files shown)


ghoneycutt:
Wow! Very impressive PR, thank you!

Could you add that Debian 8 is supported in the metadata.json file as well as the README.md ?

Are these settings the defaults that you would see with a vanilla install?

jplana:
Thank you for these great modules :-)

I just added the documentation & metadata (I completely forgot about them, sorry). The defaults that I used were taken from a vanilla install, I hope that was the correct approach.

ghoneycutt:
Yes, definitely!

ghoneycutt:
This is great. We take the approach of the vanilla install, so if you use this module you are not surprised.

ghoneycutt:
I squashed your commits into one and released them in v2.21.0

Thanks again!

PR #126, opened 2016-02-10
Added support for the minor release Debian 8.3 (Jessie)

mrodm:
This PR adds basic support for the next minor (8.3) in Debian current stable (Jessie at the moment) following the changes did with 8.2 previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ only), 1.9.3, 2.0.0 and 2.1.0.
  * Ubuntu 12.04 LTS
  * Ubuntu 14.04 LTS
  * Debian 8.2
+ * Debian 8.3
 
 EL no longer requires the `redhat-lsb` package.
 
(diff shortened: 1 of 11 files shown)


ghoneycutt:
Are there any differences between 8.2 and 8.3?

Perhaps we just need to change the matching from using `'8.2'` to `/^8/`

mrodm:
Yes, it could be changed to `/^8/`, I didn't see any differences between these two debian versions.

ghoneycutt:
Right on, are you up to sending a PR?

ghoneycutt:
Thank you!

ghoneycutt:
This is all the same as 8.2. I think we can say that the module supports Debian 8 and leave the minor versions alone. What do you think?

mrodm:
Yes, it could be done. As you say it is the same as in Debian 8.2.
In that case, what should it be done with the new tests added ?

ghoneycutt:
Appreciate all the help. This module supports Debian 8.

PR #132, opened 2016-04-25
Add param to manage nsswitch service from pam module.

Mandos:
Default behaviour is to include nsswitch module inside pam module, this
param allow to use different way of configuration nsswitch service.
Backward compaltibility was kept (default: true).

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ Boolean to purge the limits.d directory.
 
 manage_nsswitch
 ------------------
-Boolean to manage nsswitch service.
+Boolean to manage the inclusion of the nsswitch class.
 
 - *Default*: true
 
--- a/README.md
+++ b/README.md
@@ -434,6 +434,12 @@ Boolean to purge the limits.d directory.
 
 - *Default*: false
 
+manage_nsswitch
+------------------
+Boolean to manage nsswitch service.
+
+- *Default*: true
+
 ===
 
 # pam::limits::fragment define
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -53,7 +53,13 @@ class pam (
   $manage_nsswitch                     = true,
 ) {
 
-  if $manage_nsswitch == true {
+  if is_string($manage_nsswitch) == true {
+    $manage_nsswitch_real = str2bool($manage_nsswitch)
+  } else {
+    $manage_nsswitch_real = $manage_nsswitch
+  }
+
+  if $manage_nsswitch_real == true {
     include ::nsswitch
   }
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,9 +50,12 @@ class pam (
   $system_auth_ac_password_lines       = undef,
   $system_auth_ac_session_lines        = undef,
   $vas_major_version                   = '4',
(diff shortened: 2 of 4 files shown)


tehmaspc:
Need this feature as well. Thanks!

README.md
```diff
@@ -434,6 +434,12 @@ Boolean to purge the limits.d directory.
 
 - *Default*: false
 
+manage_nsswitch
+------------------
+Boolean to manage nsswitch service.
```
ghoneycutt: `Boolean to manage the inclusion of the nsswitch class.`

Mandos:
> Gem::InstallError: json_pure requires Ruby version ~> 2.0.
> Installing json 1.8.3
> An error occurred while installing json_pure (2.0.2), and Bundler cannot

@ghoneycutt Why is this happend? Is it bug in travis/rvm? I checked it locally at rbenv and bundler installed me json_pure (1.8.3).

manifests/init.pp
```diff
@@ -50,9 +50,12 @@
   $system_auth_ac_password_lines       = undef,
   $system_auth_ac_session_lines        = undef,
   $vas_major_version                   = '4',
+  $manage_nsswitch                     = true,
 ) {
```
ghoneycutt: needs validation. Puppet's eco system also has a history of turning things into strings, so if `'true'` or `'false'` get passed, we need to convert to booleans.

Check out this example

https://github.com/ghoneycutt/puppet-module-pam/blob/master/manifests/init.pp#L1097-1102

spec/classes/init_spec.rb
```diff
@@ -781,6 +781,38 @@
...
+        let :facts do
+          { :osfamily => v[:osfamily],
+            :"#{v[:releasetype]}" => v[:release],
+            :lsbdistid => v[:lsbdistid],
+          }
+        end
+        it { is_expected.to contain_class('nsswitch') }
+      end
+      
+      context "with manage_nsswitch parameter set to true" do
```
ghoneycutt: check for `'true'` and `true`. Here's an example, though you would drop the false elements from the array.

https://github.com/ghoneycutt/puppet-module-pam/blob/master/spec/classes/init_spec.rb#L784

ghoneycutt:
It needs json_pure pinned in the Gemfile like https://github.com/Ericsson/puppet-module-cron/blob/master/Gemfile#L34

I can take that on.

ghoneycutt:
I fixed the Gemfile problem with PR #140 while adding support for ruby 2.3.1. Please rebase against master to bring that in.

You can do that by doing this.

``` bash
git checkout master
git pull upstream master # provided you have my remote named 'upstream'
git checkout manage_nsswitch
git rebase master
```

Mandos:
Today I will do changes according your sugestions.

tehmaspc:
@Mandos @ghoneycutt

any chance of getting this merged in? thanks very much!

ghoneycutt:
Thank you everyone! I rebased the commits, squashing where appropriate and moving the .gitignore piece to its own commit. See PR #149

tehmaspc:
@ghoneycutt rock! thanks a bunch buddy!

PR #134, opened 2016-05-25
Add Ubuntu 16.04 LTS Support

jplana:
This change adds the latest LTS support to this module. I've tested it with a clean installation of an Ubuntu Server and it seems to work fine. I haven't tested with VAS though.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -826,39 +826,7 @@ class pam (
               $default_package_name         = 'libpam0g'
 
               if $ensure_vas == 'present' {
-                $default_pam_auth_lines = [
-                  'auth        required    pam_env.so',
-                  'auth        sufficient  pam_vas3.so show_lockout_msg get_nonvas_pass store_creds',
-                  'auth        requisite   pam_vas3.so echo_return',
-                  'auth        required    pam_unix.so use_first_pass',
-                ]
-
-                $default_pam_account_lines = [
-                  'account sufficient  pam_vas3.so',
-                  'account requisite   pam_vas3.so echo_return',
-                  'account [success=1 new_authtok_reqd=done default=ignore]  pam_unix.so',
-                  'account requisite   pam_deny.so',
-                  'account required    pam_permit.so',
-                ]
-
-                $default_pam_password_lines = [
-                  'password sufficient  pam_vas3.so',
-                  'password  requisite pam_vas3.so echo_return',
-                  'password  [success=1 default=ignore]  pam_unix.so obscure sha512',
-                  'password  requisite pam_deny.so',
-                  'password  required  pam_permit.so',
-                ]
-
-                $default_pam_session_lines = [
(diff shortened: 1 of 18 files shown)


manifests/init.pp
```diff
@@ -820,8 +820,77 @@
                 ]
               }
             }
+            '16.04': {
+              $default_pam_d_login_template = 'pam/login.ubuntu16.erb'
+              $default_pam_d_sshd_template  = 'pam/sshd.ubuntu16.erb'
+              $default_package_name         = 'libpam0g'
+
+              if $ensure_vas == 'present' {
+                $default_pam_auth_lines = [
```
ghoneycutt: If this is just copy/paste and has not been tested with VAS, then we should just fail() and say that Ubuntu 16.04 does not support VAS. I would rather do that than assume it works.

manifests/init.pp
```diff
@@ -820,8 +820,77 @@
...
+                  'session required  pam_permit.so',
+                  'session optional  pam_umask.so',
+                  'session required  pam_vas3.so create_homedir',
+                  'session requisite pam_vas3.so echo_return',
+                  'session required  pam_unix.so',
+                ]
+
+              } else {
+
+                $default_pam_auth_lines = [
```
ghoneycutt: Did you copy these values from 14.04 or from a fresh Ubuntu install?

ghoneycutt:
Wow, stellar PR! I wish they were all this nice :)

manifests/init.pp
jplana: Ok, you're right, this is untested code. I just copied it from 14.04, given that the changes in the rest of the flies are minimum between releases. Are the values from 14.04 just the defaults  after installing pam_vas? Would it be enough to just install pam_vas under 16.04 and get the defaults into the manifest? Otherwise, probably the best option is as you suggested just to fail()

manifests/init.pp
jplana: Yep, I did got them from an Ubuntu Server fresh install. Does any of them looks suspicious? I can triple checks them in any case. Changes were minimum between 14.04 and 16.04.

ghoneycutt: Right on, just checking they came from the distro and were not copied from 14.04

manifests/init.pp
ghoneycutt: VAS is a commercial product. If you do not run it, then just doing the fail() is fine. Someone that uses it can add support if needed.

http://software.dell.com/products/authentication-services/

jplana:
Thanks!

manifests/init.pp
jplana: Didn't know about VAS, and thought I could find it freely available. I'll remove the code then.

ghoneycutt:
Thanks for adding support for Ubuntu 16.04!! Apologies for the lag. I went ahead and squashed your commits in PR #136 and will be merging from there.

PR #137, opened 2016-07-19
Added more general support for Debian 8

jplana:
While updating some hosts, I realized that the current Debian Jessie (8) support was to specific, as it only supported 8.2.

I changed the way the module identifies Debian hosts so it's now possible to use the major version number. This also fixes the identification of the (not supported) Debian Wheezy hosts (7).


jplana:
Some of the errors that I see in Travis are completely unrelated, they seem to be related to this change: 251b17dd95afb9173e910831e5b570ac2a7a20ed. In my own environment I just reverted Gemfile. 

BTW It seems that puppet-lint just got a [new version]( https://rubygems.org/gems/puppet-lint/versions/2.0.0) but couldn't use it, as the puppet-lint plugins depend on version 1.0.

ghoneycutt:
Thank you!

Fixing Travis in PR #138 and rebased this in PR #139

PR #141, opened 2016-08-16
Add template for performing custom sshd pam config

boandersson:
Specific use case is to be able to supply options to pam_vas3 that are
specifically needed (and should only be configured) for sshd to resolv
intermittent ssh login failures with QAS.

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -273,6 +273,44 @@ describe 'pam' do
 
   describe 'config files' do
     platforms.sort.each do |k,v|
+      context "when configuring pam_d_sshd_template on #{v[:osfamily]} with #{v[:releasetype]} #{v[:release]}" do
+        let :facts do
+          { :osfamily => v[:osfamily],
+            :"#{v[:releasetype]}" => v[:release],
+            :lsbdistid => v[:lsbdistid],
+          }
+        end
+
+        let (:params) do
+          { :pam_d_sshd_template     => 'pam/sshd.custom.erb',
+            :pam_sshd_auth_lines     => [ 'auth_line1', 'auth_line2' ],
+            :pam_sshd_account_lines  => [ 'account_line1', 'account_line2' ],
+            :pam_sshd_session_lines  => [ 'session_line1', 'session_line2' ],
+            :pam_sshd_password_lines => [ 'password_line1', 'password_line2' ],
+          }
+        end
+
+        sshd_custom_content = <<-END.gsub(/^\s+\|/, '')
+          |# This file is being maintained by Puppet.
+          |# DO NOT EDIT
+          |#
+          |auth_line1
+          |auth_line2
+          |account_line1
+          |account_line2
+          |password_line1
+          |password_line2
+          |session_line1
+          |session_line2
+        END
+
+        if v[:osfamily] == 'Solaris'
+          it { should_not contain_file('pam_d_sshd') }
+        else
+          it { should contain_file('pam_d_sshd').with('content' => sshd_custom_content) }
+        end
+      end
+
       context "with specifying services param on #{v[:osfamily]} with #{v[:releasetype]} #{v[:release]}" do
         let :facts do
           { :osfamily => v[:osfamily],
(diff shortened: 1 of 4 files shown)


ghoneycutt:
Thanks @boandersson 

templates/sshd.custom.erb
```diff
@@ -0,0 +1,15 @@
+# This file is being maintained by Puppet.
```
ghoneycutt: we probably need some fixtures around this for the spec testing

spec/classes/init_spec.rb
```diff
@@ -794,6 +823,41 @@
...
+
+        it "should fail" do
+            expect {
+              should contain_class('pam')
+            }.to raise_error(Puppet::Error, /pam_sshd_\[auth|account|password|session\]_lines required for template pam\/sshd.custom.erb/)
+        end
+      end
+
+      [ :pam_sshd_auth_lines, :pam_sshd_account_lines, :pam_sshd_password_lines, :pam_sshd_session_lines ].each do |param|
+        context "with #{param} specified and pam_d_sshd_template not specified on #{v[:osfamily]} with #{v[:releasetype]} #{v[:release]}" do
```
ghoneycutt: what happens if multiple items are missing? will it still fail? I'm a bit skeptical of the multiple OR conditionals chained together.

spec/classes/init_spec.rb
```diff
@@ -273,6 +273,35 @@
...
+        let (:params) do
+          { :pam_d_sshd_template     => 'pam/sshd.custom.erb',
+            :pam_sshd_auth_lines     => [ 'auth_lines' ],
+            :pam_sshd_account_lines  => [ 'account_lines' ],
+            :pam_sshd_session_lines  => [ 'session_lines ' ],
+            :pam_sshd_password_lines => [ 'password_lines' ],
+          }
+        end
+
+        if v[:osfamily] != 'Solaris'
```
ghoneycutt: This condition block and the one under it should be combined into an if/else block

ghoneycutt:
@boandersson would it be possible to just update the default VAS entries in the code?

spec/classes/init_spec.rb
Phil-Friderici: Looking at the docs [1] [2] this should work as expected. I have used something similar [3] in my AFS module with no reported errors so far.

We could add spec tests to proof the expected behaviour.

[1] https://docs.puppet.com/puppet/latest/reference/lang_conditional.html#behavior
[2] https://docs.puppet.com/puppet/latest/reference/lang_expressions.html#or
[3] https://github.com/Phil-Friderici/puppet-module-afs/blob/v1.2.6/manifests/init.pp#L90-L100

templates/sshd.custom.erb
Phil-Friderici: True if the sequence does matter here. Looking at the short result of the existing tests I would recommend to use a variable instead of adding a a fixture file. Like here [1].

[1] https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/classes/init_spec.rb#L127-L149

Phil-Friderici:
An alternative approach would be to bring all sshd.$osfamily.erb template content into the code. This would become a bigger refactoring of the functionality which I would only do on a major version update.

So I can understand why @boandersson chose this way.

spec/classes/init_spec.rb
boandersson: Sure thing. The reason for doing this way in the first place was that I don't quite like conditionals in tests and thought this would be a tad more readable and obvious. Not sure what I think any more so I'll change it ;)

templates/sshd.custom.erb
boandersson: Agree that the current tests aren't that useful, esp since the sequence of each section is very important... I'll have a look.

boandersson: Took the route with a variable as suggested by @Phil-Friderici , let me know what you think

spec/classes/init_spec.rb
boandersson: Did minor updates to the tests only for the code path that actually writes the sshd file to avoid adding too much clutter. Is it sufficient or did you have something else in mind?

ghoneycutt:
I'm going to defer to @Phil-Friderici on this. When he gives the thumbs up, I'll merge and release.

spec/classes/init_spec.rb
Phil-Friderici: Here [1] is a suggestion how to test that it fails when only one of the full param set is missing.

While verifing that the test is actully doing what's intended, I realized that even if the OR chain would fail the `validate_arry()` will fail later on too.

So maybe my commit is overdoing it. Anyway, was fun to code it ;)

[1] https://github.com/boandersson/puppet-module-pam/pull/1

Phil-Friderici:
@boandersson Think we only need to agree how detailed we want to test here. After that, please squash all commits into one for a nice commit history :)

boandersson:
Added test updates from @Phil-Friderici. Let me know what you think before I squash any commits

Phil-Friderici:
From my perspective this is complete.
@boandersson please squash it into one commit.
@ghoneycutt please have a second look and merge it if you are ok with it too.

Phil-Friderici:
@boandersson thanks for squashing
@ghoneycutt looks good to me now :)

ghoneycutt:
Thank you both!

ghoneycutt:
Released in v2.27.0

boandersson:
Thanks for code review and merge @ghoneycutt !

PR #142, opened 2016-08-22
Add pam::access_lines as a fallback mechanism

jeffmccune:
Without this patch there isn't a good way to configure access rules at
different hierarchy levels, e.g. on a per-application development team basis.
Consider the situation where numerous different development teams need access
to the servers running their application.  Each team should not be allowed
access except to their own app servers.  There is an infrastructure team who
needs access to all servers.

This patch allows the infrastructure team to be listed in the
`pam::access_lines` hiera key in `common.yaml`.  App developer groups are
listed in `pam::allow_users` as normal and do not affect the rules applied to
all servers from common.yaml.

Hiera array merging is not ideal for this solution due to complexity and the
nature of allow_users potentially being a String, Array or Hash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,48 @@ This would create /etc/security/access.conf with the following content.
 + : username2 : tty1
 </pre>
 
+access_lines
+------------
+Array of String values, each string representing a complete text line in
+`/etc/security/access.conf`.  The purpose of this parameter is to configure a
+baseline set of access rules.  Entries from the `allow_users` parameter will go
+above entries specified in `allow_lines_fallback`, and the default deny-all
+entry will follow the entries specified in `allow_lines_fallback`.  Defaults to
+`['+ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6']` which implements, "User
+root should be allowed to get access via cron, X11 terminal :0, tty1, ..., tty5,
+tty6."
+
+- *Default*: ['+ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6']
+
+# Hiera example for access_lines
+<pre>
+# Allow devs to log in with a fallback of allowing ops and vagrant to log in
+# in the event pam::allowed_users gets overridden.
+pam::allowed_users:
+  - devs
+pam::access_lines:
+  - '+ : ops : ALL'
+  - '+ : vagrant : ALL'
+  - '+ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6'
+</pre>
+
+This would create /etc/security/access.conf with the following content.
+<pre>
+# This file is being maintained by Puppet.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -88,6 +88,48 @@ This would create /etc/security/access.conf with the following content.
...
+#
+
+# allow only the groups listed
++ : devs : ALL
+# pam::allowed_lines Array<String> fallback values
++ : ops : ALL
++ : vagrant : ALL
++ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6'
+
+# default deby
```
ghoneycutt: typo: 'deny'

README.md
```diff
@@ -88,6 +88,48 @@ This would create /etc/security/access.conf with the following content.
...
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+#
+
+# allow only the groups listed
++ : devs : ALL
+# pam::allowed_lines Array<String> fallback values
++ : ops : ALL
++ : vagrant : ALL
++ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6'
```
ghoneycutt: does this work? I thought that `:` is a delimiter, so expecting that `:0` would break it

README.md
```diff
@@ -88,6 +88,48 @@ This would create /etc/security/access.conf with the following content.
...
+Array of String values, each string representing a complete text line in
+`/etc/security/access.conf`.  The purpose of this parameter is to configure a
+baseline set of access rules.  Entries from the `allow_users` parameter will go
+above entries specified in `allow_lines_fallback`, and the default deny-all
+entry will follow the entries specified in `allow_lines_fallback`.  Defaults to
+`['+ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6']` which implements, "User
+root should be allowed to get access via cron, X11 terminal :0, tty1, ..., tty5,
+tty6."
+
+- *Default*: ['+ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6']
```
ghoneycutt: default should be `undef` so that we are not changing existing behavior

ghoneycutt:
Spoke with @jeffmccune and decided to punt on this functionality at this time.

README.md
jeffmccune: > > ++ : root : crond :0 tty1 tty2 tty3 tty4 tty5 tty6'
> does this work? I thought that : is a delimiter, so expecting that :0 would break it
> 

I was surprised too, but I pulled it right from the man page so presumably it works.

http://www.linux-pam.org/Linux-PAM-html/sag-pam_access.html#sag-access.conf-examples <http://www.linux-pam.org/Linux-PAM-html/sag-pam_access.html#sag-access.conf-examples>

PR #144, opened 2016-08-30
Use newer puppetlabs_spec_helper which does strict variable checking

ghoneycutt:
puppetlabs_spec_helper v1.2.0 introduced testing Puppet v4 with strict
variable checking by default.


ghoneycutt:
Released in v2.28.0

PR #147, opened 2016-09-26
Add support for Debian 7

Mandos:
Added spec tests.

### DIFF
--- a/spec/fixtures/pam_common_noninteractive_session.defaults.debian7
+++ b/spec/fixtures/pam_common_noninteractive_session.defaults.debian7
@@ -1,0 +1,6 @@
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+session [default=1]   pam_permit.so
+session requisite     pam_deny.so
+session required      pam_permit.so
+session required      pam_unix.so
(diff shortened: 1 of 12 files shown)


spec/fixtures/pam_common_noninteractive_session.defaults.debian7
```diff
@@ -0,0 +1,6 @@
+# This file is being maintained by Puppet.
+# DO NOT EDIT
+session [default=1]   pam_permit.so
```
ghoneycutt: Is all of the PAM info from a default installation? Our goal is that when you include the pam class on a freshly built system, it should not change anything. That way it is safe to use as-is without worrying if you are running a non-stock and possibly unsupported configuration.

ghoneycutt:
Big thanks for this!

spec/fixtures/pam_common_noninteractive_session.defaults.debian7
Mandos: I used vagrant's debian 7 box. I will confirm this with fresh instalation.

Mandos: Looks like this is correct default configuration.

ghoneycutt:
Thank you @Mandos !

I will release in v2.29.0 with PR #148

ghoneycutt:
Released in v2.29.0

PR #152, opened 2017-01-30
Ubuntu 16.04

dsundq:
Add support for vas

### DIFF
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -373,15 +373,6 @@ describe 'pam' do
             next
           end
 
-          if check == 'vas' and v[:osfamily] == 'Debian' and v[:release] == '16.04'
-            it 'should fail' do
-              expect {
-                should contain_class('pam')
-              }.to raise_error(Puppet::Error,/Pam: vas is not supported on #{v[:lsbdistid]} #{v[:release]}/)
-            end
-            next
-          end
-
           v[:files].each do |file|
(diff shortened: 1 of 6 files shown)


spec/classes/init_spec.rb
```diff
@@ -373,15 +373,6 @@
             next
           end
 
-          if check == 'vas' and v[:osfamily] == 'Debian' and v[:release] == '16.04'
```
ghoneycutt: this should remain, just change 16.04 to 18.04

ghoneycutt: This should remain, just change 16.04 to 18.04

ghoneycutt:
Hi, thank you for the contribution. We need a test that shows that if you have ubuntu 16.04 and vas 4 that it shows the proper files and and their content through the use of fixtures.

boandersson:
Hi @ghoneycutt, can you please clarify what you mean regarding tests? We have the tests that are included since ubuntu1604 is in the platforms hash - those tests towards the default vas_major_version (4) and are seen in the spec output: "with vas params on Debian with lsbdistrelease 16.04".
Do we need other tests as well?

boandersson:
Hi @ghoneycutt, do you have any clarification on what tests you want us to add?

ghoneycutt:
Thank you everyone! I went ahead and rebased and added a commit to support Puppet 4.9 in PR #153

PR #155, opened 2017-03-16
Support for OpenSUSE 42

MrManor:
Taking advantage on the work from @buzzdeee I more or less replicated his work and did a search/replace to support current OpenSUSE version. Briefly tested on OpenSUSE 42.2 

Note: This is my first attempt to use github so please be gentle if I miss to follow the how-to contribute correctly :-)


ghoneycutt:

PR #158, opened 2017-04-07
Restrict users by default per SSH on Debian and Ubuntu distros.

fbarbeira:
Load the module 'pam_access.so' by default on all Debian and Ubuntu
platforms because users were not restricted on SSH connections.


ghoneycutt:
I've created PR #159 to address the failing tests in travis

PR #162, opened 2017-04-24
Update PAM access to include SLES 10, 12, Ubuntu, Debian including checked to ensure they don't exist when not defined. 

benkevan:


ghoneycutt:
Awesome work!

benkevan:
Thanks. You can close the other PR since I've moved to a non-master remote branch.

.travis.yml
```diff
@@ -1,6 +1,11 @@
 ---
 language: ruby
 
+addons:
+  apt:
+    packages: 
+      - lsb-release
```
ghoneycutt: Thanks for spotting this. I think the LSB facts are a vestige of having to support Suse 9, which did not support the non-LSB facts. Not sure if that's still the case. Hopefully something we can refactor when this moves to Puppet v4 only.

.travis.yml
```diff
@@ -1,6 +1,11 @@
 ---
 language: ruby
 
+addons:
+  apt:
+    packages:
```
ghoneycutt: Seems this package is already present

```$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install lsb-release

Reading package lists...

Building dependency tree...

Reading state information...

lsb-release is already the newest version.

lsb-release set to manually installed.

0 upgraded, 0 newly installed, 0 to remove and 275 not upgraded.```

**Comment by benkevan** [comment]
I may break the pull and run Travis against my branch and do a new PR. each
step appears to find little things here and there. I've done apt packages
since Debian had been the only thing complaining. But still getting some
build failures.

**Comment by ghoneycutt** [comment]
Looks like `lsbdistid` needs to be specified in the spec tests. You can do this per check or for everything in `spec/spec_helper.rb`

**Comment by ghoneycutt** [comment]
No need for a new PR, just keep push to your branch and it will show up here.

File: spec/classes/init_spec.rb [hunk 3]
```diff
@@ -484,8 +484,10 @@
         end
         let(:params) {{ :login_pam_access => 'sufficient' }}
 
-        if (v[:osfamily] == 'RedHat') or (v[:osfamily] == 'Suse' and ['10', '11', '12'].include?(v[:release]))
+        if (v[:osfamily] == 'RedHat') or (v[:osfamily] == 'Suse' and ['10', '11', '12'].include?(v[:release])) or (v[:osfamily] == 'Debian' and ['7', '8'].include?(v[:release])) or (v[:osfamily] == 'Debian' and v[:lsbdistid] == 'Ubuntu' and ['12.04', '14.04', '16.04'].include?(v[:release]))
```
Phil-Friderici: I didn't did the math, maybe the number of else cases is smaller than this list ? Maybe turning around the else case is easier to read ?

Phil-Friderici:
@benkevan I know the pain when you just thought, let's do this quick fix.
And suddenly your realize you have opend the box of pandorras special side effects ;)

Good to see the specs fixed too. Great work !

benkevan:
@ghoneycutt should the supported pam_access tests be added to accesslogin_spec.rb rather than spec_helper.rb? Currently, the Ubuntu lsbdistid is defined within init_spec[1]. 

Overall it doesn't appear that accesslogin_spec.rb has been updated as compatibility grew.

[1] https://github.com/ghoneycutt/puppet-module-pam/blob/master/spec/classes/init_spec.rb#L165-L185

ghoneycutt:

PR #164, opened 2017-08-11
Include 'libpam-systemd' package on debian family systems.

fbarbeira:
System and service manager. Systemd cries in the journald if it is not installed:

```
PAM unable to dlopen(pam_systemd.so): /lib/security/pam_systemd.so: cannot open shared object file: No such file or directory
PAM adding faulty module: pam_systemd.so
```

Info about package:

https://packages.ubuntu.com/xenial/libpam-systemd


ghoneycutt:

PR #167, opened 2017-10-31
Drop puppet 3 support and use data types for parameters

treydock:
This PR is incomplete, I have not updated the main pam class.  I wanted to know, should things that currently accept string or boolean and use str2bool be continued or should the data type just be Boolean?  These changes likely require a major version bump so figured maybe good time to force parameters to be the proper data type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,110 +1,37 @@
 ---
 language: ruby
 
-rvm:
-  - 1.8.7
-  - 1.9.3
-  - 2.0.0
-  - 2.1.9
-  - 2.3.1
+bundler_args: --without system_tests development
 
-env:
-  matrix:
-    - PUPPET_GEM_VERSION="~> 3.1.0"
-    - PUPPET_GEM_VERSION="~> 3.2.0"
-    - PUPPET_GEM_VERSION="~> 3.3.0"
-    - PUPPET_GEM_VERSION="~> 3.4.0"
-    - PUPPET_GEM_VERSION="~> 3.5.0"
-    - PUPPET_GEM_VERSION="~> 3.6.0"
-    - PUPPET_GEM_VERSION="~> 3.7.0"
-    - PUPPET_GEM_VERSION="~> 3.8.0"
-    - PUPPET_GEM_VERSION="~> 3" FUTURE_PARSER="yes"
-    - PUPPET_GEM_VERSION="~> 4.0.0"
-    - PUPPET_GEM_VERSION="~> 4.1.0"
-    - PUPPET_GEM_VERSION="~> 4.2.0"
-    - PUPPET_GEM_VERSION="~> 4.3.0"
-    - PUPPET_GEM_VERSION="~> 4.4.0"
-    - PUPPET_GEM_VERSION="~> 4.5.0"
-    - PUPPET_GEM_VERSION="~> 4.6.0"
-    - PUPPET_GEM_VERSION="~> 4.7.0"
-    - PUPPET_GEM_VERSION="~> 4.8.0"
-    - PUPPET_GEM_VERSION="~> 4.9.0"
-    - PUPPET_GEM_VERSION="~> 4.10.0"
-    - PUPPET_GEM_VERSION="~> 4"
+cache: bundler
+
+before_install:
+  - bundle -v
+  - rm Gemfile.lock || true
+  - gem update --system
+  - gem update bundler
+  - gem --version
+  - bundle -v
 
 sudo: false
 
-script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec'
+script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec strings:generate'
 
 matrix:
   fast_finish: true
-  exclude:
-    - rvm: 2.0.0
-      env: PUPPET_GEM_VERSION="~> 3.1.0"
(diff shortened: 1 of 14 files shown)


treydock:
@ghoneycutt Just FYI, relates to PR discussion for MOTD module, the message for `Optional[Array]` changes in puppet  >= 4.10.x in such a way that the only way to match the error is just to say `/Array/`.

Puppet 4.9.x: `expects an Array value, got String`
Puppet 4.10.x/5.x: `expects a value of type Undef or Array, got String`

ghoneycutt:
Thanks for all the hard work!

I think it's time we drop string `'true', 'false'` and just use Boolean.

.travis.yml
```diff
@@ -1,110 +1,21 @@
 ---
 language: ruby
 
-rvm:
```
ghoneycutt: Let's use the same Travis config as https://github.com/sensu/sensu-puppet/blob/master/.travis.yml

Only slightly different in testing for puppet-strings and some new minor releases of Puppet 5.

manifests/limits.pp
```diff
@@ -3,31 +3,19 @@
...
-  $config_file_mode     = '0640',
-  $limits_d_dir         = '/etc/security/limits.d',
-  $limits_d_dir_mode    = '0750',
-  $purge_limits_d_dir   = false,
+  Stdlib::Absolutepath $config_file             = '/etc/security/limits.conf',
+  Optional[Array] $config_file_lines            = undef,
+  Optional[String] $config_file_source          = undef,
+  Pattern[/^[0-7]{4}$/] $config_file_mode       = '0640',
+  Stdlib::Absolutepath $limits_d_dir            = '/etc/security/limits.d',
+  Pattern[/^[0-7]{4}$/] $limits_d_dir_mode      = '0750',
```
ghoneycutt: You'd think stdlib would give us a mode pattern by now :)

metadata.json
```diff
@@ -98,6 +98,6 @@
   "dependencies": [
     {"name":"ghoneycutt/common","version_requirement":">= 1.4.1 < 2.0.0"},
     {"name":"ghoneycutt/nsswitch","version_requirement":">= 1.3.0 < 2.0.0"},
-    {"name":"puppetlabs/stdlib","version_requirement":">= 4.6.0 < 5.0.0"}
+    {"name":"puppetlabs/stdlib","version_requirement":">= 4.13.1 < 5.0.0"}
```
ghoneycutt: Curious is there something special about 4.13.1 that has functionality we are using?

ghoneycutt:
Happy to go to a new major version. Are there other things we should change about the module?

metadata.json
treydock: I think I pulled that from some official puppetlabs modules that got updated to only support Puppet 4/5.  The 4.13.0 is needed for custom Stdlib data types, like the one for validating absolute path.

.travis.yml
treydock: Updated to use the linked .travis.yml.

treydock:
@ghoneycutt Part of major version, maybe move various defaults to module level Hiera?  The pam class code is extremely long and mostly just default values.  This may be unrealistic given some defaults are based on `$ensure_vas` and `$vas_major_version`.

Maybe saving some lines by using `pick` instead of 5 line conditions for picking which value to use for `Optional` parameters.

ghoneycutt:
I'd like to drop support for VAS in the next major version and instead just document what values should be used. We could setup a vas/ directory with the hiera structure in it so that you could copy those to your hiera and have VAS working. <-- @Phil-Friderici 

Would definitely like to switch to data in modules, which we just did for sgnl05/sssd. This will also cut way down on lines of code.

https://github.com/sgnl05/sgnl05-sssd

treydock:
@ghoneycutt I'd be happy to help with such changes via different pull requests.  May be a good time to use something like https://skywinder.github.io/github-changelog-generator/ to generate CHANGELOG, something I've seen used in voxpupuli projects.

treydock:
@ghoneycutt There are several deprecated parameters in `init.pp`.  Okay to bundle a commit removing those into this pull request?

ghoneycutt:
+1 to the changelog generator. I'm doing that on other projects.

ghoneycutt:
I made a branch named puppet5. Could you point this against that branch? That way we can all work on the breaking changes and get them merged as they come in and time when we merge the whole thing into master.

I wont make any changes to master in the meantime, so we aren't stuck in rebase hell.

ghoneycutt:
Go ahead and remove the deprecated params in their own commit.

treydock:
@ghoneycutt Added commit for removing deprecated parameters, let me know if anything in this PR should be split into separate PR.

ghoneycutt:
BTW, I made https://github.com/puppetlabs/puppetlabs-stdlib/pull/834 to add Stdlib::Mode type validation.

ghoneycutt:
I dont think we actually need yard and redcarpet as puppet-strings will pull in its dependencies

Phil-Friderici:
> Happy to go to a new major version. Are there other things we should change about the module?
> @Phil-Friderici what do you think?

Changing from `'true'` and `'false'` to booleans is great.
Chaning from `'unset'` to `undef` would be great too. (used for `$pam_d_login_oracle_options`)

PR #170, opened 2017-11-25
Use Hiera 5 module data

treydock:
The hierarchy seems incorrect but was only way I could figure to maintain previous OS support behavior.


ghoneycutt:
Amazing! Thank you!

PR #172, opened 2017-12-19
Refactors

Phil-Friderici:
Using iterations with `each` instead of resource declarations for each single common* file on Linux systems.


Phil-Friderici:
As this is a quite complex refactor I have taken extra care to **NOT** change the existing spec tests.
The only change needed does adjust the internaly used resource names for $common_session_noninteractive_file on Debian/Ubuntu.

Phil-Friderici:
Added some refactoring for the common* files in https://github.com/ghoneycutt/puppet-module-pam/pull/172.

Phil-Friderici:
If you like this refactoring I will move as much data into Hiera 5 module data to clean up the code even more.

manifests/init.pp
```diff
@@ -136,324 +114,98 @@
         mode    => $pam_d_sshd_mode,
       }
 
-      case $::osfamily {
+      case $facts['os']['family'] {
```
ghoneycutt: repeating the case statement inside itself seems wrong

manifests/init.pp
```diff
@@ -136,324 +114,98 @@
...
+              fail("osfamily Suse's os.release.major is <${::facts['os']['release']['major']}> and must be 9, 10, 11, 12 or 13")
             }
           }
         }
         default: {
           fail('Pam is not supported on your osfamily')
         }
       }
+
+      $_common_files.each |$_common_file| {
```
ghoneycutt: very clever

spec/classes/init_spec.rb
```diff
@@ -268,15 +266,13 @@
         :packages           => [ 'libpam0g', ],
         :files              => [
           { :prefix         => 'pam_common_',
-            :types          => ['auth', 'account', 'password', 'session', 'noninteractive_session' ],
+            :types          => ['auth', 'account', 'password', 'session', 'session_noninteractive' ],
```
ghoneycutt: this would change the name of the file and i believe would result in both `noninteractive_session` and `session_noninteractive` being on disk. Does that pose any issue?

Phil-Friderici: I double checked the code that it only change the internal resource name but leaves the path untouched.

There was a special line (394) needed for these cases in the spec tests:
https://github.com/ghoneycutt/puppet-module-pam/blob/v2.33.0/spec/classes/init_spec.rb#L389-L403. That can now get removed again :)

manifests/init.pp
Phil-Friderici: Well, that took more time for refactoring. Checkout the commit "Refactor case statements".

Phil-Friderici:
Main class looks so much cleaner now without all theses osfamily dependend settings. Based on the great work @treydock did in https://github.com/ghoneycutt/puppet-module-pam/commit/a8e97864957089a63611ec0ac128b5b88511f729 I have now moved the rest into hiera data.

Using iteration for the common* files to get rid of the os specific resource collections.

Finalise: Use module hiera data
Fix: Get rid of lsb* facts

ghoneycutt:
Thank you!

PR #176, opened 2018-01-03
work on unit tests (please review)

Phil-Friderici:
starting to review the unit tests for completeness

- compress OS specific unit tests into one block per OS

finished:
- pam (mostly see comment below (and in unit tests))
- pam::accesslogin
- pam::limits
- pam::limits::fragment
- pam::service

/!\ untested functionality in pam
$limits_fragments_hiera_merge
$other_file
$common_auth_file
$common_auth_pc_file
$common_account_file
$common_account_pc_file
$common_password_file
$common_password_pc_file
$common_session_file
$common_session_pc_file
$common_session_noninteractive_file
$system_auth_file
$system_auth_ac_file
$password_auth_file
$password_auth_ac_file
$common_files
$common_files_create_links
$common_files_suffix

/!\ untested content validation in pam
$package_name


spec/classes/init_spec.rb
```diff
@@ -456,369 +454,311 @@
...
-            fixture = File.read(fixtures("#{filename}.defaults.#{k}"))
-            it { should contain_file(filename).with_content(fixture) }
+              if file[:symlink]
+                symlinkname = "#{file[:prefix]}#{type}"
+                symlinkpath = "#{dirpath}#{file[:prefix]}#{type}"
+                symlinkpath.gsub! '_', '-'
+                symlinkpath.sub! 'pam-', ''
+                it do
+                  should contain_file(symlinkname).with({
+                    'ensure' => 'symlink',
```
ghoneycutt: ensure would be `link` instead of `symlink`.

Phil-Friderici: Good find, I'll have to change it in the code as well.
https://github.com/ghoneycutt/puppet-module-pam/blob/puppet5/manifests/init.pp#L162

manifests/accesslogin.pp
```diff
@@ -1,18 +1,61 @@
...
-  Stdlib::Filemode $access_conf_mode      = '0644',
-  String $access_conf_template            = 'pam/access.conf.erb',
-) {
+  Stdlib::Absolutepath $access_conf_path      = '/etc/security/access.conf',
+  String $access_conf_owner                   = 'root',
+  String $access_conf_group                   = 'root',
+  Stdlib::Filemode $access_conf_mode          = '0644',
+  String $access_conf_template                = 'pam/access.conf.erb',
+  Variant[Array, Hash, String] $allowed_users = $pam::allowed_users,
+) inherits ::pam {
```
ghoneycutt: we should not be inheriting a class with parameters.

Phil-Friderici: Is there another difference when including a class via `inherits` or [`require`](https://github.com/ghoneycutt/puppet-module-pam/blob/master/manifests/accesslogin.pp#L15) beside the timing ?

- We could remove $allowed_users from main class and move it to accesslogin exclusively.

- The other way around would be something like this pseudo code:

```
class pam::accesslogin (
  Optional[Variant[Array, Hash, String]] $allowed_users = undef,
) {

case $allowed_users {
  undef: { $allowed_users_real = $pam::allowed_users }
  default: $allowed_users_real = $allowed_users }
}
```

ghoneycutt: +1 to moving `allowed_users` from `pam` to `pam::accesslogin` because the variable is only used in access.conf which is managed by `pam::accesslogin`.

Phil-Friderici: fixed

Phil-Friderici: Having issues to test package_name in accesslogin as it also comes from main class pam that is not included anymore now. Usual technique to use pre_conditions didn't worked for me today :(

ghoneycutt:
I think this all got merged into the v3 branch and released. Let me know and please re-open if that's not the case.

PR #178, opened 2018-04-30
Add support for Ubuntu 18.04

KlavsKlavsen:
Updated pam templates to match what is in 18.04.


templates/login.ubuntu18.erb
```diff
@@ -0,0 +1,19 @@
+auth     optional   pam_faildelay.so  delay=3000000
```
ghoneycutt: Is this from a stock installation without modification?

templates/sshd.ubuntu18.erb
```diff
@@ -0,0 +1,15 @@
+@include common-auth
```
ghoneycutt: Is this from a stock installation without modification?

KlavsKlavsen: Yes. there wasn't any big changes.. except for the reorganizing of selinux part.

KlavsKlavsen:
Travis test seems to fail due to something unrelated to my code :(

nbentoumi:
@ghoneycutt  @KlavsKlavsen ,
i did a test on ubuntu 18.04, don't see any issue.
:+1:

gavmain:
Hi @KlavsKlavsen, I appreciate the work you've done and I'm eager to see it pulled. It looks like this maybe failing the CI tests due to missing [spec fixtures](https://github.com/ghoneycutt/puppet-module-pam/tree/master/spec/fixtures) for ubuntu1804.

KlavsKlavsen:
@gavmain I've tried to do so - by looking at the commit that added Ubuntu 16.04 support.. it fails with an odd failure, where I can't seem to figure out what I actually did wrong.. Can you help me figure it out? (if I want to run the fixtures spec on my own end.. Have you got any hint as to what to setup on an ubuntu to be able to do that?) -- and perhaps that could be put in README.. as "Developers documentation" or something :)

gavmain:
I was a bit worried whether VAS was compatible with Ubuntu 18.04, but checking their [site](https://www.oneidentity.com/products/authentication-services/#Specifications) they support all current Ubuntu releases.

If that is the case, I think you are safe to update the version on this [line](https://github.com/KlavsKlavsen/puppet-module-pam/blob/874c292db88a0a531732588b8fad0d318f55cb33/spec/classes/init_spec.rb#L387), from:
`if check == 'vas' and v[:osfamily] == 'Debian' and v[:release] == '18.04'`
to:
`if check == 'vas' and v[:osfamily] == 'Debian' and v[:release] == '20.04'`
and see if that now passes the CI tests. 

I'm basing this assumption on the [commit](https://github.com/ghoneycutt/puppet-module-pam/commit/670653b78d4098a4fd3762e30905b0f3e5df2af1#diff-a446e4110d88d59f153d0721dad568feR376) by Daniel Sundqvist - who is no longer on Github by the looks of it - where he added VAS support for 16.04.

KlavsKlavsen:
Changed that - and the test fails the same :(

templates/sshd.ubuntu18.erb
gavmain: @KlavsKlavsen, can you add update the template to reflect the [addition](https://github.com/ghoneycutt/puppet-module-pam/pull/160/commits/32d27a53c2824b7c2b619950361815917a84bece#diff-e3cdb1800313bb4fcca541886851296bR3) that was done by @fbarbeira for sshd.ubuntu18.erb?

gavmain:
I'm just having a look. I wonder if @ghoneycutt can cast some light on this?

nbentoumi:
VAS is compatible with Ubuntu 18.04, performed a test and working fine

gavmain:
The Gemfile looks like it needs updating, specifically this [line](https://github.com/ghoneycutt/puppet-module-pam/blob/master/Gemfile#L29) relating to the metadata-json-lint. It requires Ruby 2.0 or higher according to the metadata-json-lint [readme](https://github.com/voxpupuli/metadata-json-lint#compatibility). For the job [#609.1](https://travis-ci.org/ghoneycutt/puppet-module-pam/builds/385671904) I found a similar error reported to the [metadata-json-lint project](https://github.com/voxpupuli/metadata-json-lint/issues/74).

It does look like you are facing a Ruby dependency nightmare. Digging further I see @ghoneycutt has a branch called [v3](https://github.com/ghoneycutt/puppet-module-pam/tree/v3) which has a significant number of changes to the build files and has dropped a number of old versions. I wonder if it is best to merge your work into there. We'll hopefully hear back from @ghoneycutt on what to do.

KlavsKlavsen:
But does errors must have been there before then.. I have changed nothing that should cause this.. Unless the build system is done so it just fetches "the latest" version of dependencies it uses - and I happened to make this patch - when the latest versions broke the test system :(

KlavsKlavsen:
@gavmain I've merged the same change as @fbarbeira - to sshd.ubuntu18.erb

ghoneycutt:

PR #181, opened 2018-05-15
Added lookup() function to replace hiera_hash() function

robert4man:
We are in the process of migrating from Puppet3->Puppet4 and have been refactoring our use of hiera_hash() function calls. This seems applicable here as well.


ghoneycutt:

This has been implemented in v3.

PR #185, opened 2018-07-17
Adding a define that manages generic access files for other services …

pelacables:
…than access

In case you want custom access files for other services:
/etc/pam.d/postgresql content : 
`[...]
account required pam_access.so accessfile=/etc/security/postgresql_access.conf
[...]`
this define lets you create the access file.


ghoneycutt:

PR #186, opened 2018-07-24
Add suport for Ubuntu 18.04

amateo:
This PR is to provide support for Ubuntu 18.04. By default, it leaves pam.d files with the default configuration provided by ubuntu, except for `sshd` where is added `pam_access`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -908,8 +908,84 @@ class pam (
                 ]
               }
             }
+            '18.04': {
+              $default_pam_d_login_template = 'pam/login.ubuntu18.erb'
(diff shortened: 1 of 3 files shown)


amateo:
Hi @ghoneycutt,

 a few months ago I made this PR. It failed with problems not directly related with my commit (I guess, I'm not familiar with tests).

  Should I fix something?

ghoneycutt:
Hi @amateo I'm working on releases v3 which is a major upgrade and needs 18.04 support. Could you target the `v3` branch please?

ghoneycutt:
It should take copying and pasting info into `data/os/Ubuntu/18.04.yaml` and adding fact info to `spec/spec_helper.rb`.

ghoneycutt:

PR #188, opened 2018-08-09
Add support for Debian 9

hdep:
Hello,

here is a PR for adding support for Debian 9.
Please let me know if it is ok. Currently working great on my system.

Hugo

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -976,13 +976,16 @@ class pam (
                 'session optional      pam_systemd.so',
               ]
             }
-        }
-    }
-          default: {
-              fail("Pam is only supported on Debian 7 and 8. Your lsbmajdistrelease is <${::lsbmajdistrelease}>.")
+            default: {
+              fail("Pam is only supported on Debian 7, 8 and 9. Your lsbmajdistrelease is <${::lsbmajdistrelease}>.")
             }
-      }
+          }
         }
+        default: {
+          fail("Pam is only supported on lsbdistid Ubuntu or Debian of the Debian osfamily. Your lsbdistid is <${::lsbdistid}>.")
+        } 
+      }
+    }
     'Solaris': {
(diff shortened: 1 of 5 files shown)


hdep:
Hello, I don't understand why the build failed, some are OK other are not, any help appreciated @ghoneycutt

manifests/init.pp
```diff
@@ -916,42 +916,68 @@
...
           case $::lsbmajdistrelease {
             /(7|8)/: {
-
               if $ensure_vas == 'present' {
                 fail("Pam: vas is not supported on ${::osfamily} ${::lsbmajdistrelease}")
               }
               $default_pam_d_login_template = "pam/login.debian${::lsbmajdistrelease}.erb"
               $default_pam_d_sshd_template  = "pam/sshd.debian${::lsbmajdistrelease}.erb"
               $default_package_name         = 'libpam0g'
-
```
ghoneycutt: These lines should stay for readability

templates/login.debian9.erb
```diff
@@ -0,0 +1,19 @@
+auth       optional   pam_faildelay.so  delay=3000000
```
ghoneycutt: is this data for a default, vanilla install of debian9?

hdep: I just checked on another VM, yes it does.

manifests/init.pp
hdep: Fixed in last commit

hdep:
Hello @ghoneycutt can I do something else for this PR ?

hdep:
Hi @ghoneycutt  do you still support this module ?

ghoneycutt:
You can help us by reviewing the v3 branch of this repo against any Debian and Ubuntu systems you might have.

ghoneycutt:

PR #191, opened 2018-10-20
Add acceptance tests

treydock:
The coverage is lacking, still some tests need to be added.


treydock:
Added some extra tests, can squash once everything looks good.

ghoneycutt:
Thank you!

PR #192, opened 2018-11-12
Add support por Ubuntu Bionic 18.04

amateo:
This is a rework of #186, updated to puppet 5 and with platform configuration in data/os/Ubuntu/18.04.yaml


ghoneycutt:
Awesome work, thank you! Seems to have just one failed spec test at https://travis-ci.org/ghoneycutt/puppet-module-pam/jobs/453880410.

We also need acceptance testing, for that you can just copy `spec/acceptance/nodesets/ubuntu-1604.yml` to `spec/acceptance/nodesets/ubuntu-1804.yml` and search replace 16 to 18.

For the tests to happen, they just need two entries added to `.travis.yml`.

Here's the lines for Ubuntu 16.04 LTS
https://github.com/ghoneycutt/puppet-module-pam/blob/master/.travis.yml#L75-L86

amateo:
My last push, adding entries at .travis.yml has broken something. Could you check it?

amateo:
I have made the new PR #193. This one could be closed and ignored.

PR #196, opened 2018-12-04
Synatptis fix pam password auth ac

synaptis:
MD5 is old and no longer considered secure. This PR replaces MD5 with SHA512 for RHEL 6 & 7 systems.


ghoneycutt:
@synaptis Thanks for the PR! We want to be sure that this patch does not break older systems that might be doing md5 by default. Could you link to documentation around this from RedHat? Does it require any special pam packages to be installed?

ghoneycutt:
BTW, the spec tests failed because the fixtures need to be updated.

https://github.com/ghoneycutt/puppet-module-pam/tree/master/spec/fixtures

synaptis:
> BTW, the spec tests failed because the fixtures need to be updated.
> 
> https://github.com/ghoneycutt/puppet-module-pam/tree/master/spec/fixtures

Sorry about that, I'm still a bit new to unit tests. I'll get it sorted out.

ghoneycutt:
@synaptis Thanks, this looks great! Could you please rebase and squash your commits into one commit with a subject similar to

```
(GH-196) Use SHA512 instead of MD5 for passwords on EL6 and EL7
```

I'll get this verified on some VM's and get it merged.

synaptis:
All done. Rebase and squash was new to me but I'm pretty sure it's done right.

ghoneycutt:
Released in v3.2.0

Issue #211, opened 2019-05-22
RHEL8 support


term1n4l:
Could you update this module to allow support for RHEL8?

ghoneycutt:
Please send a pull request. Here's a recent example of SLES 15 being added.

https://github.com/ghoneycutt/puppet-module-pam/commit/f87a27a9d86981efd1a93b553df1c3f511b28ea7

term1n4l:
> Please send a pull request. Here's a recent example of SLES 15 being added.
> 
> [f87a27a](https://github.com/ghoneycutt/puppet-module-pam/commit/f87a27a9d86981efd1a93b553df1c3f511b28ea7)

Seems Pete Harvey beat me to it, thanks Pete Harvey :+1:

marszip:
When do expect to have a formal release that includes this fix?

ghoneycutt:
Released in v3.4.0 !

PR #213, opened 2019-05-24
(GH-211) Add support for RHEL 8

ph84172:
I've tested this on a fresh install of RHEL 8 and it appears to be working correctly.

Garrett, I'm not sure how you want to handle the vagrant image for this test as there's no CentOS 8 yet.  There does appear to be a vagrant box for RHEL 8 but I haven't tested it.


ghoneycutt:
Hi @ph84172 could you also please add a vagrant box to the Vagrantfile so we can test it works that way. We also need to add a docker image to https://github.com/ghoneycutt/puppet-module-pam/tree/master/spec/acceptance/nodesets and https://github.com/ghoneycutt/puppet-module-pam/blob/master/.travis.yml which I forgot to do for SLES15.

ph84172:
@ghoneycutt is there a way to specify a different container registry in the nodesets file?  Redhat publish their "Universal Base Image" which is based on RHEL8 but it's hosted on registry.access.redhat.com rather than Docker Hub.

ph84172:
Alternatively we wait until CentOS 8 has been released and there's a centos8 image on Docker Hub :)

ghoneycutt:
Try specifying the registry through the environment variable `DOCKER_REGISTRY` such as

```sh
DOCKER_REGISTRY="registry.access.redhat.com" \
BEAKER_set="el-8" \
BEAKER_PUPPET_COLLECTION=puppet6 \
bundle exec rake beaker
```

ghoneycutt:
If that does not work, suggest using one of the unofficial images from docker hub.

ghoneycutt:
Created https://tickets.puppetlabs.com/browse/BKR-1593

ghoneycutt:
Hi @ph84172 Seems we just have these two left before we can merge. For the docker image, suggest using an unofficial one and putting a comment in there with a link to this issue that explains why it is being used and that we are waiting on the BKR ticket to use the official one.

- [ ] Vagrant
- [ ] beaker nodeset and addition to .travis.yml

ph84172:
@ghoneycutt I've tried adding unofficial RHEL 8 Vagrant boxes / Docker images - does this run through Travis correctly?

Vagrantfile
```diff
@@ -21,6 +21,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider :virtualbox do |vb|
     vb.customize ["modifyvm", :id, "--memory", "512"]
   end
+  
+  config.vm.define "el8-pam", autostart: true do |c|
```
ghoneycutt: autostart should be false. When you do `vagrant up` without specifying a VM it will start those and we only want one to start, EL7.

spec/acceptance/nodesets/rhel-8.yml
```diff
@@ -0,0 +1,20 @@
...
+    docker_image_commands:
+      - 'yum install -y wget'
+    docker_container_name: 'pam-el8'
+CONFIG:
+  log_level: debug
+  type: foss
+ssh:
+  password: root
+  auth_methods: ["password"]
+
```
ghoneycutt: no empty line at the bottom

spec/acceptance/nodesets/rhel-8.yml
```diff
@@ -0,0 +1,20 @@
+HOSTS:
+  rhel8:
+    roles:
+      - agent
+    platform: el-8-x86_64
+    hypervisor: docker
+    image: roboxes/rhel8
```
ghoneycutt: could you add a comment here that points to the BKR ticket that explains we are not using an official image because we cant until that ticket is resolved.

.travis.yml
```diff
@@ -48,6 +48,18 @@ matrix:
     env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6
     bundler_args:
     script: bundle exec rake beaker
+  - rvm: 2.4.4
```
ghoneycutt: perfect

ph84172:
It's all yours, thanks Garrett.

Might be worth thinking about switching the box/container images over to CentOS 8 when it's released to keep it consistent with the previous use of CentOS 7 as the canonical "EL7" target.

ghoneycutt:
Docker and Vagrant failing because you need credentials to use yum. Can we work around this?

```
"This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register."
```

ph84172:
I've removed the "yum -y install wget" in the nodeset file as wget is already installed in the base RHEL 8 environment.

I also reworked provision_basic_el.sh into a new provision_basic_rhel.sh for the RHEL 8 target which again, removes the wget reference and removes references to EPEL (doesn't exist yet for RHEL 8) and CentOS repositories.

ph84172:
I've also renamed the Vagrant machine from "el8-pam" to "rhel8-pam" thinking that it's probably better to start distinguishing explicitly between CentOS (el) and RHEL (rhel).

ph84172:
Ping @ghoneycutt - how's this PR looking now?  Thanks.

ghoneycutt:
@ph84172 Could you please rebase and I'll get this merged. Great work on this PR!

ph84172:
@ghoneycutt looks like Beaker tries to do a "yum install" of some openssh components once the container is up, which will fail without a valid RH subscription.  Given that we can't be too far away from a CentOS 8 release I'm wondering if we should nix this until then?  

We could either add support for EL 8 without functioning tests for now and put it on the "unsupported" list or just wait for CentOS 8 and try this PR again then.  What do you think?

ghoneycutt:
Seems like a reasonable approach to put them in unsupported, pending centos 8 release. Besides the README, could you update travis with the an allowed_failures section and put a comment that these should be changed to centos 8 and link back to this pull request. That should hopefully make it apparent to everyone what's going on and when to expect support.

here's an example

https://github.com/sensu/sensu-puppet/blob/master/.travis.yml#L140-L144

ph84172:
@ghoneycutt thanks - now updated.

spec/acceptance/nodesets/centos-8.yml
```diff
@@ -0,0 +1,20 @@
+HOSTS:
+  centos7:
```
ghoneycutt: this should read `centos8`

spec/acceptance/nodesets/centos-8.yml
```diff
@@ -0,0 +1,20 @@
...
+    docker_image_commands:
+      - 'yum install -y wget'
+    docker_container_name: 'pam-el8'
+CONFIG:
+  log_level: debug
+  type: foss
+ssh:
+  password: root
+  auth_methods: ["password"]
+
```
ghoneycutt: no empty line

ghoneycutt:
@ph84172 Drats! Noticed a typo with centos7 vs 8. Could you please update/rebase and then I'll get this merged for real :)

ph84172:
@ghoneycutt should be good to go 👍

ph84172:
Ping @ghoneycutt 😊

anders-larsson:
HI @ghoneycutt,

Is there anything preventing this from being merged? We would love to see RHEL8 support added :)

anders-larsson:
Hi again,

After some further investigation is appears we might need to do further changes for this pull request. In RHEL8 they've introduced [authselect](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/using_authselect_on_a_red_hat_enterprise_linux_host/index) which appears to be a new way of customizing PAM configuration. It allows you to have multiple PAM configuration and select which PAM configuration should be used.

Current implemention (old way) still appears to work however maybe we should try to align us to RedHat's way of working? This could make it easier to introduce non-default setups as well such as QAS.

ph84172:
It's certainly worth revisiting - the current PR had some workarounds for the fact there were no CentOS 8 images around at the time.

ghoneycutt:
Thanks everyone!

PR #217, opened 2019-11-16
Add support for Debian 10

thechristschn:
There ist currently no package for a puppet5 puppetserver on Debian 10, so I added travis tests only for puppet6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,12 @@ matrix:
     env: BEAKER_set="debian-9" BEAKER_PUPPET_COLLECTION=puppet6
     bundler_args:
     script: bundle exec rake beaker
+  - rvm: 2.4.4
+    sudo: required
+    services: docker
+    env: BEAKER_set="debian-10" BEAKER_PUPPET_COLLECTION=puppet5
+    bundler_args:
+    script: bundle exec rake beaker
   - rvm: 2.5.3
     sudo: required
     services: docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,12 @@ matrix:
     env: BEAKER_set="debian-9" BEAKER_PUPPET_COLLECTION=puppet6
     bundler_args:
     script: bundle exec rake beaker
+  - rvm: 2.5.3
+    sudo: required
+    services: docker
+    env: BEAKER_set="debian-10" BEAKER_PUPPET_COLLECTION=puppet6
+    bundler_args:
+    script: bundle exec rake beaker
   - rvm: 2.4.4
     sudo: required
     services: docker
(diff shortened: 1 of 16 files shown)


ghoneycutt:
Thank you for the contribution!

.travis.yml
```diff
@@ -84,6 +84,12 @@ matrix:
     env: BEAKER_set="debian-9" BEAKER_PUPPET_COLLECTION=puppet6
     bundler_args:
     script: bundle exec rake beaker
+  - rvm: 2.5.3
```
ghoneycutt: please add puppet 5

templates/sshd.debian10.erb
```diff
@@ -0,0 +1,18 @@
...
+session    optional     pam_keyinit.so force revoke
+@include common-session
+session    optional     pam_motd.so  motd=/run/motd.dynamic
+session    optional     pam_motd.so noupdate
+session    optional     pam_mail.so standard noenv # [1]
+session    required     pam_limits.so
+session    required     pam_env.so # [1]
+session    required     pam_env.so user_readenv=1 envfile=/etc/default/locale
+session [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so open
+
```
ghoneycutt: please remove trailing blank line

ghoneycutt:
Are these configs from a standard install without any other configuration added?

.travis.yml
thechristschn: With which rvm version? There is no puppetserver5 package for debian 10 available.

templates/sshd.debian10.erb
thechristschn: Fixed.

thechristschn:
> Are these configs from a standard install without any other configuration added?

This is the default config from a new standard installation. The only change I made was to remove `@include common-password` from the end of `templates/sshd.debian10.erb` . But this line was already removed for debian9 from the standard installation, so I guess there is a reason. But I can readd it if you want.

.travis.yml
ghoneycutt: It doesn't use the puppetserver, it uses the agent package.

ghoneycutt:
Please add that line back so we can keep it as default as possible. Thank you!

thechristschn:
> Please add that line back so we can keep it as default as possible. Thank you!

Done.

ghoneycutt:
Vagrant is failing

```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

apt-get install -y linux-headers-`uname -r` dkms

Stdout from the command:

Reading package lists...
Building dependency tree...
Reading state information...


Stderr from the command:

E: Unable to locate package linux-headers-4.19.0-5-amd64
E: Couldn't find any package by glob 'linux-headers-4.19.0-5-amd64'
E: Couldn't find any package by regex 'linux-headers-4.19.0-5-amd64'
```

ghoneycutt:
Vagrant is working. Great work!

PR #221, opened 2020-05-05
Add support for Ubuntu 20.04

runejuhl:
This PR adds support for Ubuntu 20.04 (focal fossa).

I've based the PAM defaults on the current `ubuntu:20.04` Docker image; there should only be minor differences compared to Ubuntu 18.04.

I had an issue with conflicting dependies on `stdlib` versions and opted to simply bump the version ranges. There might be a better fix, but ultimately it's your decision if you want to allow users to use older `stdlib` versions.

PR #224, opened 2020-06-23
Focal

amateo:
Add support for Ubuntu 20.04


ghoneycutt:
Hi @amateo check out the work done in #221 

That is pretty close to being ready, perhaps you can base your work off that pull request and add the small things that are left.

PR #233, opened 2021-01-18
adding Amazon Linux OS support

lduriez:
It should work as RedHat OS family but don't enter in the if statement you put, and I relly need this cause my production server are under Amazon Linux 2 OS


lduriez:
Didn't see there was another PR about Amazon Linux 2, I close, sorry

PR #238, opened 2021-07-19
Add config option for session-noninteractive

promasu:


promasu:
Just stumbled across this PR. Is the new commit what you meant by needs tests?
I did not manage to find a test for the other pam_FOOBAR_lines parameters

ghoneycutt:
Thank you for your contribution. Looks like you can just add your data to `pam_session_lines` and then there is no need for another parameter. This also keeps the module simpler and you know all the lines come from the one array instead of them occurring in two.

promasu:
Thanks for looking into this.
Yes you can do that and this stays the default with this PR, but there are usecases where you want different PAM-Lines for interactive and non-interactive sessions. This is why I opened this PR in the first place

PR #239, opened 2021-09-22
Added support for Debian 11

ZloeSabo:
This PR adds support for Debian 11 "Bullseye". 

* I've taken the settings for `pam::pam_*_lines:` from the default system config, which is slightly different from Debian 10.
* There is no puppet 5 build for Debian 11, so I've only added 6 to the Travis config. The module does not officially support puppet 7, so I didn't add a configuration for it.
* Tested with `BEAKER_set="debian-11" BEAKER_PUPPET_COLLECTION=puppet6 bundle exec rspec spec/acceptance/` and with a vagrant box
* Also added JUnit output to the gitignore


ghoneycutt:
Very well done, thank you!

johnmharris:
Great timing for us, thank you.

PR #241, opened 2021-10-01
Missing entries in pam.d/sshd for Ubuntu 16

gnobironts:
The /etc/pam.d/sshd file generated for Ubuntu 16.04 are missing these two lines compared to default:

```bash
# Print the status of the user's mailbox upon successful login.
session    optional     pam_mail.so standard noenv # [1]

# Read environment variables from /etc/environment and
# /etc/security/pam_env.conf.
session    required     pam_env.so # [1]
```

This PR fixes this.

Templates for Ubuntu 18 and Ubuntu 20 are already ok.

### DIFF
--- a/templates/sshd.ubuntu16.erb
+++ b/templates/sshd.ubuntu16.erb
@@ -11,6 +11,7 @@ session  optional     pam_keyinit.so force revoke
 session  optional     pam_motd.so  motd=/run/motd.dynamic
 session  optional     pam_motd.so  noupdate
 session  required     pam_limits.so
+session  required     pam_env.so # [1]
 session  required     pam_env.so user_readenv=1 envfile=/etc/default/locale
 session  [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so open
 @include common-password
--- a/templates/sshd.ubuntu16.erb
+++ b/templates/sshd.ubuntu16.erb
@@ -10,6 +10,7 @@ session  optional     pam_keyinit.so force revoke
 @include common-session
 session  optional     pam_motd.so  motd=/run/motd.dynamic
 session  optional     pam_motd.so  noupdate
+session  optional     pam_mail.so standard noenv # [1]
 session  required     pam_limits.so
 session  required     pam_env.so # [1]
 session  required     pam_env.so user_readenv=1 envfile=/etc/default/locale
--- a/spec/fixtures/pam_d_sshd.defaults.ubuntu1604
+++ b/spec/fixtures/pam_d_sshd.defaults.ubuntu1604
@@ -8,7 +8,9 @@ session  optional     pam_keyinit.so force revoke
 @include common-session
 session  optional     pam_motd.so  motd=/run/motd.dynamic
 session  optional     pam_motd.so  noupdate
+session  optional     pam_mail.so standard noenv # [1]
 session  required     pam_limits.so
+session  required     pam_env.so # [1]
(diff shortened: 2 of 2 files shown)


templates/sshd.ubuntu16.erb
```diff
@@ -10,7 +10,9 @@ session  optional     pam_keyinit.so force revoke
 @include common-session
 session  optional     pam_motd.so  motd=/run/motd.dynamic
 session  optional     pam_motd.so  noupdate
+session  optional     pam_mail.so standard noenv # [1]
```
ghoneycutt, line 13: To confirm, these are from a default install?

gnobironts: Yes,
On a Default Ubuntu 16.04 install I have the following content :
```
cat sshd |grep -v  "^#" |grep -v "^$"
@include common-auth
account    required     pam_nologin.so
@include common-account
session [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so close
session    required     pam_loginuid.so
session    optional     pam_keyinit.so force revoke
@include common-session
session    optional     pam_motd.so  motd=/run/motd.dynamic
session    optional     pam_motd.so noupdate
session    optional     pam_mail.so standard noenv # [1]
session    required     pam_limits.so
session    required     pam_env.so # [1]
session    required     pam_env.so user_readenv=1 envfile=/etc/default/locale
session [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so open
@include common-password
```

And with the comments:
```
# PAM configuration for the Secure Shell service

# Standard Un*x authentication.
@include common-auth

# Disallow non-root logins when /etc/nologin exists.
account    required     pam_nologin.so

# Uncomment and edit /etc/security/access.conf if you need to set complex
# access limits that are hard to express in sshd_config.
# account  required     pam_access.so

# Standard Un*x authorization.
@include common-account

# SELinux needs to be the first session rule.  This ensures that any
# lingering context has been cleared.  Without this it is possible that a
# module could execute code in the wrong domain.
session [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so close

# Set the loginuid process attribute.
session    required     pam_loginuid.so

# Create a new session keyring.
session    optional     pam_keyinit.so force revoke

# Standard Un*x session setup and teardown.
@include common-session

# Print the message of the day upon successful login.
# This includes a dynamically generated part from /run/motd.dynamic
# and a static (admin-editable) part from /etc/motd.
session    optional     pam_motd.so  motd=/run/motd.dynamic
session    optional     pam_motd.so noupdate

# Print the status of the user's mailbox upon successful login.
session    optional     pam_mail.so standard noenv # [1]

# Set up user limits from /etc/security/limits.conf.
session    required     pam_limits.so

# Read environment variables from /etc/environment and
# /etc/security/pam_env.conf.
session    required     pam_env.so # [1]
# In Debian 4.0 (etch), locale-related environment variables were moved to
# /etc/default/locale, so read that as well.
session    required     pam_env.so user_readenv=1 envfile=/etc/default/locale

# SELinux needs to intervene at login time to ensure that the process starts
# in the proper default security context.  Only sessions which are intended
# to run in the user's context should be run after this.
session [success=ok ignore=ignore module_unknown=ignore default=bad]        pam_selinux.so open

# Standard Un*x password updating.
@include common-password
```

ghoneycutt:
Thanks! These options likely came in a later release from when this was made.

PR #243, opened 2021-12-15
Convert to use PDK and restructure spec tests

anders-larsson:
Questions:
* Should we remove CI files that are not being used (configure deletion in .sync.yaml)
* Move to GitHub Actions for automation/CI instead of travis? If we want to continue to use travis it needs to be updated to use travis-ci.com instead?)
* How do we handle the acceptance tests? Seems like PDK does not support it natively but does refer to Litmus https://puppetlabs.github.io/litmus/

TODO:
* ~Fix remaining warnings from PDK validate~
* ~Convert spec tests to use *on_supported_os.each(?)~

BTW: Removed validation tests for parameters. Not sure if we want to keep them or not. If we want to keep them I can remove the commit.


spec/classes/accesslogin_spec.rb
```diff
@@ -20,66 +20,76 @@
     END
 
     it do
-      should contain_file('access_conf').with({
-        'ensure'  => 'file',
+      is_expected.to contain_file('access_conf').with({
+                                                        'ensure' => 'file',
```
anders-larsson, line 24: I've seen this happen before with pdk validate is executed. Will fix this and other occurrences.

anders-larsson:
Made a temp commit with the changes to get rid of:
```
pdk (CONVENTION): rubocop: Don't repeat descriptions within an example group. (spec/acceptance/00_pam_spec.rb:31:11)
pdk (CONVENTION): rubocop: Don't repeat descriptions within an example group. (spec/acceptance/00_pam_spec.rb:33:11)
pdk (CONVENTION): rubocop: Don't repeat descriptions within an example group. (spec/acceptance/00_pam_spec.rb:44:9)
pdk (CONVENTION): rubocop: Don't repeat descriptions within an example group. (spec/acceptance/00_pam_spec.rb:46:9)
```

Not sure if it is still working though. Need to be verified.

spec/classes/init_spec.rb
```diff
@@ -120,99 +121,97 @@
...
-      let(:params) {{ :login_pam_access => 'sufficient' }}
+      let(:params) { { login_pam_access: 'sufficient' } }
 
-      case v[:pam_d_login]
-      when 'without_pam_access'
-        it { should contain_file('pam_d_login').without_content(/account[\s]+sufficient[\s]+pam_access.so/) }
-      when 'with_pam_access'
-        it { should contain_file('pam_d_login').with_content(/^account[\s]+sufficient[\s]+pam_access.so$/) }
+      if v[:pam_d_login] == 'without_pam_access'
+        it { is_expected.to contain_file('pam_d_login').without_content(%r{account[\s]+sufficient[\s]+pam_access.so}) }
```
anders-larsson, line 189: `pdk (CONVENTION): rubocop: Empty example group detected. (spec/classes/init_spec.rb:184:5)` by converting case Change statement to if to get rid of this warning. Example group is not empty for any OS. Maybe a bug report should be made to rubocp about it?

spec/classes/init_spec.rb
```diff
@@ -120,99 +121,97 @@
...
-      let(:params) {{ :sshd_pam_access => 'sufficient' }}
+      let(:params) { { sshd_pam_access: 'sufficient' } }
 
-      case v[:pam_d_sshd]
-      when 'without_pam_access'
-        it { should contain_file('pam_d_sshd').without_content(/account[\s]+sufficient[\s]+pam_access.so/) }
-      when 'with_pam_access'
-        it { should contain_file('pam_d_sshd').with_content(/^account[\s]+sufficient[\s]+pam_access.so$/) }
+      if v[:pam_d_sshd] == 'without_pam_access'
+        it { is_expected.to contain_file('pam_d_sshd').without_content(%r{account[\s]+sufficient[\s]+pam_access.so}) }
```
anders-larsson, line 202: pdk (CONVENTION): rubocop: Empty example group detected. (spec/classes/init_spec.rb:198:5)
Change statement to if to get rid of this warning. Example group is not empty for any OS. Maybe a bug report should be made to rubocp about it?

anders-larsson:
```
1) pam on debian-10-x86_64 is expected to contain File[pam_common_auth] with ensure => "file", path => "/etc/pam.d/common-auth", content  supplied string, owner => "root"
, group => "root" and mode => "0644"
     Failure/Error:
       is_expected.to contain_file(filename).with(
         'ensure'  => 'file',
         'path'    => path,
         'content' => File.read(fixtures("#{filename}.defaults.#{platform_identifier}")),
         'owner'   => 'root',
         'group'   => group,
         'mode'    => '0644',
       )

       expected that the catalogue would contain File[pam_common_auth] with content set to supplied string
       Diff:
       @@ -1,6 +1,6 @@
        # This file is being maintained by Puppet.
        # DO NOT EDIT
        auth  [success=1 default=ignore]  pam_unix.so nullok_secure                                                                                                                -auth  requisite     pam_deny.so
       -auth  required      pam_permit.so
       +auth  requisite   pam_deny.so
       +auth  required    pam_permit.so

     # ./spec/classes/init_spec.rb:62:in `block (6 levels) in <top (required)>'

Finished in 25 minutes 3 seconds (files took 1 minute 33.06 seconds to load)
3159 examples, 1 failure
```

I really don't understand why this error occurs. Because looking at the code in Hiera and comparing it to the contents in the fixtures it looks fine. However for whatever reason extra spaces are added after `auth`. Any suggestion? We could of course fix it by just updating the fixture to include these extra spaces, but it doesn't make sense that this happens.

---

We probably want to write the whole access hash data logic. Wanted to keep it as is initially to ensure data is not lost. Any suggestions on how to solve this is welcome. Off the top if my head we could:
* Rename hash keys to same value as is used in on_supported_os
* Lookup using os_facts[:os]['name'] and then os_facts[:os]['family'] allowing us to use the same hash for all RedHat family OSs. If we for whatever reason need to diverge we can add a key to the hash structure with osname instead
* Add filename suffix as a key within said hash and use it for fixtures

What do you think @ghoneycutt ?

ghoneycutt:
I'm out of office for a bit. Let's see what @Phil-Friderici says about this.

Phil-Friderici:
> * Should we remove CI files that are not being used (configure deletion in .sync.yaml)
> 
> * Move to GitHub Actions for automation/CI instead of travis? If we want to continue to use travis it needs to be updated to use travis-ci.com instead?)

yes please do so

Phil-Friderici:
> * Rename hash keys to same value as is used in on_supported_os

Doing so will reduce brainfuck, yes please :)

Phil-Friderici:
> > * Rename hash keys to same value as is used in on_supported_os
> 
> Doing so will reduce brainfuck, yes please :)

see PR https://github.com/anders-larsson/puppet-module-pam/pull/3 for my work on that topic

anders-larsson:
We still need to decide what to do with OpenSuSE 13.1 which seems to be unsupported by its vendor and when I ran spec tests it failed due to being unsupported by the nsswitch class (included).

Should it be kept or removed?

ghoneycutt:
SuSE support is only there because of your needs, so if you and @Phil-Friderici no longer need it, let's remove it.

anders-larsson:
OK. Removed OpenSuSE from metadata.json.

How should we handle beaker? Beaker configuration was removed by PDK. I think otherwise this pull request should be in a good shape now or what do you say @Phil-Friderici ?

Phil-Friderici:
I agree that the PR should be in a good shape now. Beside the downside of Beaker being removed it adds Github Actions that run PDK based validation and unit tests.

anders-larsson:
@ghoneycutt What is the status of this? Anything missing?

Ubuntu 22.04 has been released. I've prepared a pull request for it but it depends on the changes made in this pull request.

Thanks

ghoneycutt:
@anders-larsson I'm not seeing where github actions is running.

PR #245, opened 2022-05-05
Support Ubuntu 22.04

anders-larsson:
From my testing this works correctly. Very similar configuration to 20.04. Only difference between files installed during installation and the configuration files in Puppet is that `account  required     pam_access.so` is added to the sshd PAM config file.

Anyone know if this line should be there or not? It exists on 20.04 too but not sure if it is an addition in Puppet or was in the original file.

WIP reason: facterdb does not contain facts for Ubuntu-22.04 and no spec tests are actually executed for said OS.


anders-larsson:
Convert it to draft to ensure it's not merged without proper testing.

anders-larsson:
Formatted the YAML configuration in data/os/Ubuntu/22.04.yaml to ensure there are no changes to the config lines in the PAM configuration files. Only comments are removed/updated.

However another change I found is that `session optional        pam_systemd.so` gets added to /etc/pam.d/common-session-noninteractive because it is using the same configuration data as /etc/pam.d/common-session where it is present.

ashish1099:
any update on this one ?

anders-larsson:
Removed WIP. We've been using this code for quite a while now without any glaring issues.

anders-larsson:
I see that there are a few yaml values with whitespaces before string ends. Should probably fix that. Also update README .

anders-larsson:
Making some fixes. Please wait with merging it.

anders-larsson:
Should be OK now. Had to rebase it too.

Aman1994:
How far away are we from getting this in production?  We need this

treydock:
@anders-larsson I think the default is to not enable pam_access.  Spinning up a fresh `ubuntu:22.04` container and installing `openssh-server` I see this:

```
# Uncomment and edit /etc/security/access.conf if you need to set complex
# access limits that are hard to express in sshd_config.
# account  required     pam_access.so
```

Also some improvements have been made to the testing of this module, can you do a rebase against master?

anders-larsson:
Will fix this tomorrow morning!

ashish1099:
everything is good now, can we please get this merged now ?
Thanks

ghoneycutt:
Thank you for your efforts!

Released in v3.10.0

PR #242, opened 2021-12-01
Add support for EL9

anders-larsson:
Adds support for EL9-based operating systems.

Would it be OK to convert the module to use PDK to make spec testing easier?

RHEL9 is currently in beta and centos9 container does not exist AFAIK.

PR #249, opened 2022-08-16
Fix PAM password/system file on EL8

anders-larsson:
Fixes incorrect PAM configuration for EL8 introduced by https://github.com/ghoneycutt/puppet-module-pam/commit/81560cd19cfda5091fb3036c6438bc3195fef3fb. Assumption was that the data in fixtures was correct, which was not the case. 

After this change we're using the exact configuration after EL8 installation for system-auth and password-auth.

```
# diff -y --suppress-common-lines  /etc/pam.d/system-auth-{ac,orig}
# This file is being maintained by Puppet.                    | # Generated by authselect on Tue Aug 16 10:34:46 2022
# DO NOT EDIT                                                 | # Do not modify this file manually.
# Auth                                                        |
# Account                                                     <
# Password                                                    <
# Session                                                     <
# diff -y --suppress-common-lines  /etc/pam.d/password-auth-{ac,orig}                                                                                                                                                
# This file is being maintained by Puppet.                    | # Generated by authselect on Tue Aug 16 10:34:46 2022
# DO NOT EDIT                                                 | # Do not modify this file manually.
#                                                             |
# Auth                                                        <
# Account                                                     <
# Password                                                    <
# Session                                                     <
```


treydock:
@anders-larsson This appears to assume SSSD is enabled which is not the default for minimal EL8 environment.

Example defaults from fresh container:

```
# cat /etc/pam.d/password-auth
#%PAM-1.0
# This file is auto-generated.
# User changes will be destroyed the next time authselect is run.
auth        required      pam_env.so
auth        sufficient    pam_unix.so try_first_pass nullok
auth        required      pam_deny.so

account     required      pam_unix.so

password    requisite     pam_pwquality.so try_first_pass local_users_only retry=3 authtok_type=
password    sufficient    pam_unix.so try_first_pass use_authtok nullok sha512 shadow
password    required      pam_deny.so

session     optional      pam_keyinit.so revoke
session     required      pam_limits.so
-session     optional      pam_systemd.so
session     [success=1 default=ignore] pam_succeed_if.so service in crond quiet use_uid
session     required      pam_unix.so
```

anders-larsson:
That could indeed be the case. Thanks for pointing it out. Will close this as it does not seem to be relevant.

I suspect that #242 should be updated as well. I'll look into it.

anders-larsson:
Oh. Just realised this change included more changes than the sssd one. I'll check it tomorrow what is relevant in a minimal installation.

ghoneycutt:
@anders-larsson @treydock is this ok to be closed or is the fix needed?

anders-larsson:
Well.. It seems to work anyway so maybe it's not worth the effort to make changes retroactively (even though they're more aligned with the defaults). Might cause problems for users I suspect.

Will close it.

PR #251, opened 2022-12-15
Add Ubuntu 22.04 support

kolaCZek:


ghoneycutt:
There are differences between #254 and #251. The default values should match a default, minimal install of the platform. Which values are correct?

treydock:
This is superseded by #245

ghoneycutt:
Thank you for your efforts!

Released in v3.10.0

PR #256, opened 2023-01-01
Add support for EL9

ghoneycutt:


ghoneycutt:
This is a rebase of #242

anders-larsson:
Hi @ghoneycutt,

We might need to update this to not include sssd PAM module. Haven't verified yet but if we base the module on "minimal" the case might be that it should not be included. I'll have a look tomorrow.

ghoneycutt:
Hi @anders-larsson I'm not familiar with EL9. Does it no longer use sssd by default like EL8?

anders-larsson:
I'll have to verify it tomorrow. I'll make a minimal install and check whether sssd is installed by default or not. According to comment in #249 sssd was not default in minimal install in EL8.

anders-larsson:
Just to make sure. Should be module be aligned with a minimal install of RHEL or a "base" install?

bschonec:
I just created a new RHEL9 based on "minimal-environment" and the following sssd packages were installed:

```
sssd-client-2.6.2-2.el9.x86_64
sssd-common-2.6.2-2.el9.x86_64
sssd-kcm-2.6.2-2.el9.x86_64
```

anders-larsson:
Thanks for verifying.

In that case the correct configuration should be valid.

PR #260, opened 2023-05-23
Add parameter to control manamgent of access.conf

imp-:
Add a flag to control the inclusion of the `pam::accesslogin` class.
This allows the module to be used when access.conf is already managed externally.


ghoneycutt:
Thank you for your contribution @imp- ! This functionality has been added in v4.1.0.

PR #266, opened 2023-07-17
Add pam::limits::purge_limits_d_dir_ignore parameter

### LINKED ISSUE #226: Support for a limits purge ignore flag
I'd like to see support for an ignore parameter which would _not_ delete files matching the glob in the limits directory.

We use Ansible for maintaining some application level configuration on our hosts. While we can turn off the `purge_limits_d_dir` flag we actually want Puppet to own the directory. However, we also need the ability to occasionally write some application specific overrides.

treydock:
This parameter will allow the limits.d purge to ignore specified files during purge Fixes #226


manifests/limits.pp
```diff
@@ -33,6 +36,7 @@
   Stdlib::Absolutepath $limits_d_dir   = '/etc/security/limits.d',
   Stdlib::Filemode $limits_d_dir_mode  = '0750',
   Boolean $purge_limits_d_dir          = false,
+  Optional[Variant[String[1], Array]] $purge_limits_d_dir_ignore = undef,
```
ghoneycutt, line 39: Perhaps this should be array of non empty strings instead of just an array.

treydock: This has been updated to require an array of non-empty strings.

ghoneycutt (approved):

PR #268, opened 2023-08-17
Added noop_value as param so that access_conf file can be updated wit…

Aman1994:
…h noop false too

### DIFF
--- a/manifests/accesslogin.pp
+++ b/manifests/accesslogin.pp
@@ -16,6 +16,9 @@
 # @param access_conf_mode
 #   Mode of access.conf.
 #
+# @param noop_value
+#   Noop value to write access_conf file.
+#
 # @param access_conf_template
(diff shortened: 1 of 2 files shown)


Aman1994:
This change is required since we have 100 servers that we manage and there is a need more often than not to add a user in minimum 10 servers - so, we have to go and run Puppet in no-noop on each server which is very tedious.  But, with this change we can easily deploy the change in noop without any manual intervention

ghoneycutt:
You could achieve this in a profile using resource overrides. See https://www.puppet.com/docs/puppet/7/lang_resources.html#adding-or-modifying-attributes

```puppet
profile::maintenance_window {

  if $allowed_to_modify_stuff == false {
    File['access_conf'] {
      noop = true,
    }
  }
}
```

PR #269, opened 2023-11-14
Support Puppet 8 and stdlib 9.x

treydock:


treydock:
This won't work until https://github.com/trlinkin/puppet-nsswitch/pull/49 is merged and released.

PR #270, opened 2024-03-11
Support managing faillock.conf and pwquality.conf

treydock:


manifests/faillock.pp
```diff
@@ -0,0 +1,75 @@
...
+  String[1] $config_file_group = 'root',
+  Stdlib::Filemode $config_file_mode = '0644',
+  String[1] $config_file_template = 'pam/faillock.conf.erb',
+  Optional[Stdlib::Filesource] $config_file_source = undef,
+  Stdlib::Absolutepath $dir = '/var/run/faillock',
+  Optional[Boolean] $audit_enabled = undef,
+  Optional[Boolean] $silent = undef,
+  Optional[Boolean] $no_log_info = undef,
+  Optional[Boolean] $local_users_only = undef,
+  Integer $deny = 3,
```
ghoneycutt, line 50: Can we put min/max limits here? Do they accept 0, negative?

ghoneycutt: Integer[0]

treydock: This is done

spec/classes/faillock_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+            audit_enabled: true,
+            silent: true,
+            no_log_info: true,
+            local_users_only: true,
+            deny: 1,
+            fail_interval: 2,
+            unlock_time: 3,
+            even_deny_root: true,
+            root_unlock_time: 4,
+            admin_group: 'admins'
```
ghoneycutt, line 75: missing trailing `,`

treydock: This is fixed.

